PR #2, opened 2020-02-18
Parallel vec ops

vb000:


yodada (commented):
Thank you so much for the PR! Looks good to me! I'll add a few more tests for in-place operations.

cbatten:
Do we have tests? Ideally all pull requests should have tests ... which pass on travis. This is one reason we really want the emulation layer. We need a way to test everything with CI and make sure it works before approving PRs ...

cbatten:
Also please include a comment about what the pull request does when you submit it. Thanks!

yodada:
Yes. Tests for mul was included in the PR.

Sincerely,
Lin

cbatten:
where are the tests? without the emulation layer there is no way to run the tests in CI which means they are not useful for approving PRs.

before merging in PRs we need a process by which we will manage such PRs.

yodada:
Tests live here hb_device/torch/tests/test_hammerblade_kernels.py

vb000:
Hi, Sorry about that... was not sure hb-device is the main protected branch with ci.

Will follow the usual PR process from next time.

PR #3, opened 2020-02-18
Change the file structure of HammerBlade tensorlib

yodada:
In this PR, all code that runs on HammerBlade (tensorlib) is moved to pytorch/hammerblade.

In-place add / sub / mul tests are added.
It also fixes a few typos and minor issues.

Proof of correctness from cosim

> ================== 22 passed, 1 warning in 280.97s (0:04:40) ===================

This PR also implements single title emulation layer.


aten/src/ATen/native/hammerblade/BinaryOpsKernel.cpp
```diff
@@ -25,7 +25,7 @@ void mul_kernel_hb(TensorIterator& iter) {
   if (iter.dtype() == ScalarType::Float) {
     offload_op_binary(iter, 1.0, "tensorlib_mul");
   } else {
-    AT_ERROR("HammerBlade only supports adding two floats");
+    AT_ERROR("HammerBlade only supports multiplying two floats");
```
vb000, line 28: I was wondering if we can use a function definition macro to factor repetitive code...

vb000 (dismissed):

aten/src/ATen/native/hammerblade/BinaryOpsKernel.cpp
cbatten: I would like to avoid over using macros. At least until we know more about what we want to do. We can potentially use real C++ metaprogramming to help avoid boiler plate code.

vb000: I agree, macro are be used sparingly. In this particular scenario though, we need functions with very similar definitions and different _names_. No sure if there's a C++ construct to parameterize function names... this seems to be fundamentally a preprocessing problem.

yodada:
Eventually we would like to get rid of the Makefile system and use CMake to build the kernel

vb000:
As I understand pytorch cosim execution flow more, I feel there is a way to avoid cosim related details in pytorch altogether.

Since pytorch cosim is essentially executing `test_loader.cosim` executable instead of `python` executable, all we need `test_loader.cosim` executable, which is actaully built in `bsg_bladerunner/bsg_replicant` already. Like we depend on `libbsg_manycore_runtime.so` in `bsg_bladerunner`, we could depend on `test_loader.cosim`. This decouples cosim from PyTorch!

PR #6, opened 2020-03-04
Tensor Manipulations

vb000:
- Turn off debug feature by default (~3x speedup in cosim!)
- Division kernel
- Tensor views
- Tensor resizing


hammerblade/torch/kernel/kernel.cpp
```diff
@@ -90,6 +90,35 @@ extern "C" {
     return rc;
   }
 
+  int  __attribute__ ((noinline)) tensorlib_div(
```
yodada, line 93: The kernel function author needs to register the functions so we can emulate them in Travis ;) See line 142 as an example.
`HB_EMUL_REG_KERNEL(tensorlib_add, bsg_tensor_t*, bsg_tensor_t*, bsg_tensor_t*, float*);`

hammerblade/torch/tests/test_hammerblade_div.py
```diff
@@ -0,0 +1,50 @@
...
+
+Bandhav Veluri
+"""
+
+import torch
+
+def test_elementwise_div_1():
+    x = torch.ones(1, 10)
+    y = torch.ones(1, 10)
+    z = x * y
```
yodada: Oops

hammerblade/torch/tests/test_hammerblade_div.py
```diff
@@ -0,0 +1,50 @@
...
+Tensor division tests
+
+Bandhav Veluri
+"""
+
+import torch
+
+def test_elementwise_div_1():
+    x = torch.ones(1, 10)
+    y = torch.ones(1, 10)
```
yodada, line 11: Testing dividing two `ones` tensor seems not that useful though ...

hammerblade/torch/kernel/kernel.cpp
vb000: Right, I realized it after the CI failed -_-.

hammerblade/torch/tests/test_hammerblade_div.py
vb000: Thanks!

hammerblade/torch/tests/test_hammerblade_div.py
vb000: Dividing ones is more of an indexing test for division kernel than a test for the actual division.

aten/src/ATen/native/hammerblade/Resize.h
```diff
@@ -0,0 +1,85 @@
+#pragma once
+
+#include <ATen/ATen.h>
+#include <TH/THTensor.hpp>
+#include <ATen/hammerblade/HammerBladeContext.h>
+#include <ATen/native/hammerblade/Offload.h>
+
+namespace at { namespace native {
+
+static inline void HBStorage_resize(THStorage* storage, ptrdiff_t size) {
```
yodada, line 10: We need to think about this more actually ... THStorage is legacy storage code for CPU. This implementation definitely works and I understand you implemented it this way since CPU/CUDA they do so, but I was wondering if we should stay away from TH* in our HammerBlade code.

I think `THTensor*` is the same as `TensorImpl*` and `THStorage*` is just `StorageImpl*`.

yodada: Another option is we keep it like this in the PR, and I'll refactor it after merging

vb000: Nice suggestion! True, I didn't realize the transition to legacy objects is unnecessary here!

yodada (approved):
I think it looks good to be merged. Thanks for implementing these functions!

A note to myself: we would like to refactor `HBStorage_resize` a little bit to remove `TH*` later.

PR #8, opened 2020-03-05
De-refactor kernel functions

yodada:
Included in this PR:
 - Each kernel function lives in its own cpp file
 - Removed the lambda function we used for element-wise operations, since we want to start with the simplest implementation
 - Remove tests/target.py  any file that starts with `test_` in tests dir will be run by pytest
 - Changed the way we interface with emulation layer in kernel source files. We always include `hammerblade_emul.hpp`
 - Removed legacy interface from resize logic.
 - Fixed a potential bug in hammerblade_resize in which when resizing to 0, the tensor no longer lives on hammerblade.


vb000:
@yodada, a thought on lambda functions... the objective was two fold:

- Reduces boilerplate code in kernel development.
- Simplifies  parallelization in the long run -- even if we implement single thread kernels now, using lambda function enables us to implement parallelization primitives in one function but impact lots of kernels!

cbatten:
@vb000 we should motivate refactoring. We start by writing the code and then after we have written many kernels we can figure out the best way to do refactoring. We should avoid premature refactoring. There might be other ways which are more effective. To start we should write super clean, simple (possibly a bit too verbose) kernels ... then we can take a step back to focus on refactoring and abstractions.

aten/src/ATen/native/hammerblade/Resize.h
```diff
@@ -1,34 +1,38 @@
...
+static inline void hammerblade_resize(c10::StorageImpl* storage, ptrdiff_t size) {
   if (storage->resizable()) {
     /* case when the allocator does not have a realloc defined */
+
+    int device = c10::hammerblade::current_device();
     at::DataPtr new_data;
     if (size != 0) {
       new_data = storage->allocator()->allocate(storage->itemsize() * size);
+    } else {
+      new_data = at::DataPtr(nullptr, at::Device(at::DeviceType::HAMMERBLADE, device));
```
yodada, line 19: This is kind of important. Even if a tensor has 0 size, we still need to make sure it lives on hammerblade.

vb000: I might be missing something here... if the size is 0, there's nothing related this tensor on the device, right? And, the metadata `tensor.device` would be on the host, correct?

vb000 (approved):
Looks good!

aten/src/ATen/native/hammerblade/Resize.h
yodada: You are right in terms that if the size is 0, there is nothing on device. However, we still want to mark this tensor as a HammerBlade tensor. The way to do so is to have a HammerBlade DataPtr. ;)

PR #9, opened 2020-03-09
Conv2d operator

vb000:
- Conv2d kernel for HB.
- Tensor wrapper classes for on device tensors to better access data.
- Added `-fuse-ld=gold` to use gold linker to reduce link time. This works with devtoolset-8 which is what this repo supports.


vb000:
Superseded by https://github.com/cornell-brg/hb-pytorch/pull/12

PR #10, opened 2020-03-11
Preparing for MNIST

yodada:
This PR is kind of big ... I'll definitely switch to do PR's more frequently.

Included in this PR:
 - define `eva_t` as `uint64_t` when doing emulation
 - files for `addmm (mm)`, `threshold`, and `sigmoid`. Related tests are skipped for now. Undergrads/Meng in brg is currently working on these kernels.
 - added `HammerBladeGenerator`
 - re-organized HammerBlade host/device code a little bit ... so mostly each kernel lives in its own files.
 - `abs`, `dot`, and `fill_` kernels are implemented. A few helper functions are added to `Copy.h/cpp` to make offloading happen.
 - enable `.hammerblade` for `torch.nn.Module`
 - enable direct printing in python. Now you can pass a HammerBlade tensor to `print()`
 - `strides` in `TensorIterators` are actually important. In certain cases it could be 0, meaning we are adding each element with a single scalar.


README.md
```diff
@@ -39,6 +52,28 @@ This work aims to port PyTorch to HammerBlade.
...
+use_c10_dispatcher: full
+supports_named_tensor: True
+variants: function, method
+dispatch:
+  CPU: sigmoid
+  CUDA: sigmoid
++ HammerBlade: sigmoid
+  MkldnnCPU: mkldnn_sigmoid
+ ```
+ 2. Add host code to `aten/src/ATen/native/hammerblade/Sigmoid.cpp` Add the dummiest host code possible, without calling the kernel.
```
vb000, line 71: Adding a check that fails, makes sense here? Like `TORCH_CHECK(false, "Sigmoid not implemented");`.

README.md
```diff
@@ -39,6 +52,28 @@ This work aims to port PyTorch to HammerBlade.
...
+ 3. Add tests to `hammerblade/torch/tests/test_sigmoid.py`
+ 4. With Emulation Layer, make sure the code compiles and tests fail only because of incorrect results
+ 5. Add kernel code to `hammerblade/torch/kernel/kernel_sigmoid.cpp`, which is also the dummiest code.
+ 6. Change the host code to be more realistic: call the kernel and do nothing else.
+ 7. Implement both the host and kernel code for real, assuming 1x1 tile group.
+ 8. Make sure everything pass on Emulation layer, and write more tests. Then you are ready to create a PR!
+ 9. Make sure your code works on COSIM.
```
vb000, line 78: This currently involves rebuilding pytorch. A todo that we could look into in the future...

aten/src/ATen/HammerBladeGenerator.h
```diff
@@ -0,0 +1,35 @@
+#pragma once
+
+#include <ATen/core/Generator.h>
+#include <ATen/core/MT19937RNGEngine.h>
+
+namespace at {
+
+struct CAFFE2_API HammerBladeGenerator : public Generator {
```
vb000, line 8: Just curious, which kernel needed this random number generator?

README.md
```diff
@@ -21,11 +21,24 @@ This work aims to port PyTorch to HammerBlade.
...
+    `python3.6 -m venv ./venv_pytorch`
+    `python3.6 -m venv ./venv_pytorch`
+ - Install dependencies
+    `pip install numpy pyyaml mkl mkl-include setuptools cmake cffi typing sklearn tqdm pytest`
+ - Init pytorch third party dependencies
+    `git submodule update --init --recursive`
+ - Setup building environment variables.
+    `cd hb-pytorch && source setup_emul_build_env.sh`
+ - Build pytorch. This step can take up to 15 minutes
+    `python setup.py install`
```
vb000: A minor thing... `python setup.py develop` is faster for incremental builds. `develop` only creates symlinks for python modules whereas `install` copies them.

aten/src/ATen/native/hammerblade/Offload.cpp
```diff
@@ -36,23 +36,20 @@ static eva_t create_device_tensor(uint32_t N, uint32_t dims,
...
   }
+  // populate local_strides
+  for(int i=0; i<dims; i++) {
+    local_strides[i] = (uint32_t)strides[i];
+  }
+  // copy strides
+  dst = (void *) ((intptr_t) tensor_strides);
+  src = (void *) ((intptr_t) local_strides);
+  c10::hammerblade::memcpy_host_to_device(dst, src, dims * sizeof(uint32_t));
+  free(local_strides);
```
vb000, line 52: I like this! Copying even output strides simplifies the kernel code.

aten/src/ATen/native/hammerblade/Offload.cpp
```diff
@@ -85,11 +82,97 @@ static void cleanup_device(std::vector<eva_t> args, std::vector<eva_t> ptrs) {
...
+void offload_iterator_op_impl(TensorIterator& iter, Scalar alpha,
+    const char* kernel, uint32_t ntensors) {
 
   TORCH_INTERNAL_ASSERT(iter.can_use_32bit_indexing());
-  TORCH_INTERNAL_ASSERT(iter.ntensors() == 3); // output, input1, and input2
+  TORCH_INTERNAL_ASSERT(iter.ntensors() == ntensors);
+
+  // It is very important to use serial_for_each here, since we assume a single
+  // HammerBlade device in the system
+  iter.serial_for_each([&](char** data, const int64_t* strides, int64_t n) {
```
vb000, line 140: I believe it's okay to use just `for_each` here, because it doesn't assume multiple devices but only multiple host threads. And, we disable host multi-threading by the repo level parameter `USE_OPENMP=0` as `libbsg_manycore_runtime.so` is not thread-safe yet.

aten/src/ATen/native/hammerblade/Offload.h
```diff
@@ -23,10 +23,71 @@ typedef struct {
...
+
+#define HB_OFFLOAD_TENSOR_KERNEL_5ARGS(arg0, arg1, arg2, arg3, kernel)                                  \
+do {                                                                                                    \
+  std::vector<Tensor> args;                                                                             \
+  args.push_back(arg0);                                                                                 \
+  args.push_back(arg1);                                                                                 \
+  args.push_back(arg2);                                                                                 \
+  args.push_back(arg3);                                                                                 \
+  std::vector<Scalar> scalars;                                                                          \
+  offload_tensor_scalar_impl(args, scalars, kernel);                                                    \
```
vb000, line 89: I'm missing something here... how do we populate scalar arguments?

c10/hammerblade/emul/kernel_trampoline.cpp
```diff
@@ -1,11 +1,11 @@
 #include <kernel_trampoline.h>
 
-std::map<std::string, std::function<int(uint32_t, uint32_t*)>> kernelMap;
-std::vector<std::function<int(uint32_t, uint32_t*)>> enqueued_kernel;
+std::map<std::string, std::function<int(uint32_t, uint64_t*)>> kernelMap;
+std::vector<std::function<int(uint32_t, uint64_t*)>> enqueued_kernel;
 std::vector<uint32_t>  enqueued_argc;
-std::vector<uint32_t*> enqueued_argv;
+std::vector<uint64_t*> enqueued_argv;
```
vb000, line 6: Doesn't this increase the chance of mismatch between cosim and emulation? If it makes it easy to maintain emulation code, it makes sense though... just curious.

torch/_tensor_str.py
```diff
@@ -206,6 +206,8 @@ def _tensor_str(self, indent):
     summarize = self.numel() > PRINT_OPTS.threshold
     if self.dtype is torch.float16 or self.dtype is torch.bfloat16:
         self = self.float()
+    if self.is_hammerblade:
+        self = self.cpu()
```
vb000, line 210: Smart! =) How does cuda do it? `.float()`?

aten/src/ATen/HammerBladeGenerator.h
vb000: Never mind, found this below. For initializing nn model parameters?

vb000 (approved):

aten/src/ATen/HammerBladeGenerator.h
yodada: actually the first use case is in `nn.Dropout`, which calls `bernoulli_`

README.md
yodada: good point! I'll switch to use `develop`

aten/src/ATen/native/hammerblade/Offload.cpp
yodada: I was playing safe ;) since I noticed that pytorch still links against pthreads

torch/_tensor_str.py
yodada: Well I don't really know about the cuda way actually. We don't have cuda installed on our server so it won't compile XD

c10/hammerblade/emul/kernel_trampoline.cpp
yodada: Yes your concern is more than valid. The issue I was facing is that, it is really hard to request a large chunk of low address memory (>2M), even with `mmap`. If we want to run MNIST then we certainly needs a buffer larger than that. So I changed the definition of `bsg_tensor_t` as well, with a macro to choose between `uint32_t` and `uint64_t`.

README.md
yodada: Absolutely! I'm adding it now

aten/src/ATen/native/hammerblade/Offload.h
yodada: These marcos are for the kernels that use tensor arguments only, and thus scalars are empty.
I was trying to use the same offloading function (`offload_tensor_scalar_impl`), which handles arbitrary number of tensors followed by  arbitrary number of scalars.

yodada: I think we might need to start splitting the `Offload.h/cpp` into multiple files. It is growing fast.

vb000: Yes, I felt that too when adding conv offloading. 

I believe we can have `Offload.h` include multiple offload related headers so that we could do `#include Offload.h` to access all offloading routines -- like we do now.

vb000: > These marcos are for the kernels that use tensor arguments only, and thus scalars are empty.
> I was trying to use the same offloading function (`offload_tensor_scalar_impl`), which handles arbitrary number of tensors followed by arbitrary number of scalars.

Variadic arguments?

vb000: I think we could do 

```
template <typename... T1, typename... T2>
offload_tensor_scalar_impl(T1... tensors, T2... scalars, char* kernel) {
}
```

We don't need to template types here... but when I tried this, I was seeing compilation error without templated types.

yodada: ah good point. Let's merge this one and your conv2d in, and then let's think about what should be our offloading interface. I guess we need to make some decisions so we won't keep changing things

PR #12, opened 2020-03-11
Conv2D operator - conflicts resolved

yodada:
- Conv2d kernel for HB.
- Tensor wrapper classes for on device tensors to better access data.
- Added `-fuse-ld=gold` to use gold linker to reduce link time. This works with devtoolset-8 which is what this repo supports.


vb000:
Looks good to me :p

Let me know your comments...

aten/src/ATen/TensorUtils.cpp
```diff
@@ -132,6 +132,31 @@ void checkAllSameGPU(CheckedFrom c, ArrayRef<TensorArg> tensors) {
...
+    if (! t2->is_hammerblade()) {
+      oss << "Tensor for " << t2 << " is not on HB, ";
+    }
+    oss << "but expected " 
+        << ((!(t1->is_hammerblade() || t2->is_hammerblade())) ? "them" : "it")
+        << " to be on HB (while checking arguments for " << c << ")";
+    AT_ERROR(oss.str());
+  }
+  TORCH_CHECK(
+    t1->get_device() == t2->get_device(),
```
yodada, line 150: Currently HammerBlade is hardcoded to have device index == -1. But I think there is no harm to check it, for extensibility.

aten/src/ATen/native/hammerblade/Offload.cpp
```diff
@@ -54,6 +54,50 @@ static eva_t create_device_tensor(uint32_t N, uint32_t dims,
   return tensor;
 }
 
+static eva_t create_device_tensor(const Tensor& tensor,
```
yodada, line 57: This is awesome

hammerblade/torch/kernel/bsg_tensor.hpp
```diff
@@ -13,4 +31,100 @@ typedef struct {
...
+// =========================================================
+// Device Tensor classes
+//
+// Wrapper classes around device tensor structs to provide
+// convenience operations. This runs on a tiny RISC-V processor 
+// on the device, so be careful about using dynamic memory 
+// allocation.
+// =========================================================
+
+class BSGTensor {
```
yodada, line 52: I really like the idea of helper classes!

hammerblade/torch/kernel/bsg_tensor.hpp
```diff
@@ -13,4 +31,100 @@ typedef struct {
 #endif
 } bsg_tensor_t;
```
yodada, line 32: Actually, I think I will add a new data structure for offloading a TensorIterator.

hammerblade/torch/kernel/bsg_tensor.hpp
vb000: Yes, with operator overloading, writing device kernels feels like writing numpy!

hammerblade/torch/kernel/bsg_tensor.hpp
```diff
@@ -13,4 +31,100 @@ typedef struct {
...
+    uint32_t dim(uint32_t d) {
+      if(d >= dims) {
+        bsg_printf("BSGTensor error: dimesnion must be less than %d\n",
+            dims);
+      }
+
+      uint32_t dim;
+
+      if(d == 0) {
+        dim =  N / strides[0];
```
yodada, line 79: > Actually, I think I will add a new data structure for offloading a TensorIterator.

The reason is that for tensors constructed out of TensorIterators, `strides[i]` can be 0. If we don't careful enough we will have a floating point exception here. So I think I'll add a new set of structs/classes just for iterators.

setup_cosim_build_env.sh
```diff
@@ -15,6 +15,7 @@ export USE_QNNPACK=0
 export USE_DISTRIBUTED=0
 export USE_OPENMP=0
 export ATEN_THREADING=NATIVE
+export CFLAGS='-fuse-ld=gold'
```
yodada, line 18: Could you fill us in on why we need to use `gold`?

hammerblade/torch/kernel/bsg_tensor.hpp
vb000: Sounds good!

Couple of things I understood about TensorIterators:
- Those are exclusively for element-wise operations and hence are always 1-d tensors.
- Strides means number of bytes probably to distinguish float vs bfloat16 on GPUs.

setup_cosim_build_env.sh
vb000: As you might have experienced, linking can be a bottleneck. `gold` is a faster linker available in devtoolset-8 and is compatible with `ld`.

aten/src/ATen/native/Convolution.cpp
```diff
@@ -644,6 +644,23 @@ at::Tensor _convolution(
...
+  } else if (input.device().type() == c10::DeviceType::HAMMERBLADE) {
+    TORCH_CHECK(input.type() == weight.type(),
+             "Input type (", input.type().toString(), ") and weight type (", weight.type().toString(),
+             ") should be the same");
+    TORCH_CHECK(!bias.defined() || (input.type() == bias.type()),
+             "Input type (", input.type().toString(), ") and bias type (", bias.type().toString(),
+             ") should be the same");
+    
+    if (params.transposed) {
+      output = at::hb_convolution_transpose(
```
yodada, line 656: It seems to me that we can directly call `at::native::hb_convolution_transpose` instead of going through dynamic dispatching?

aten/src/ATen/native/native_functions.yaml
```diff
@@ -1802,6 +1802,14 @@
   dispatch:
     CUDA: miopen_depthwise_convolution_backward_weight
 
+- func: hb_convolution_transpose(Tensor self, Tensor weight, Tensor? bias, int[] padding, int[] output_padding, int[] stride, int[] dilation, int groups) -> Tensor
```
yodada, line 1805: > It seems to me that we can directly call `at::native::hb_convolution_transpose` instead of going through dynamic dispatching?

Are we preparing for having a sparse version? If no then I guess we don't need dispatching here, since only HammerBlade calls this?

hammerblade/torch/kernel/bsg_tensor.hpp
cbatten: Definitely! Developing these kind of C++ wrappers will really help kernel evelopment.

setup_cosim_build_env.sh
yodada: Cool!

aten/src/ATen/native/native_functions.yaml
vb000: This is something I don't understand fully. I tried to follow cudnn here:
https://github.com/cornell-brg/hb-pytorch/blob/cf58f27867492f137d97aaec547f33c7f54540c3/aten/src/ATen/native/native_functions.yaml#L790-L794

And all other calls in this file are made to `at::` namespace, so didn't choose to directly invoke the `at::native` here...

yodada: emmm... I looked at git blame and it shows Edward Z. Yang added this ... he must have some reason for doing so ... let's follow the cudnn style then

vb000: I just verified directly calling ` at::native::hb_convolution_transpose` works. 

But no call in that file is made to `at::native::` namespace. What do you think? Would you like me to change it?

vb000:
Good to merge?

aten/src/ATen/native/native_functions.yaml
yodada: Personally I think it would be nice to stay the same as cuDNN ;)

vb000 (approved):
Lin couldn't approve as he created this pr... so I'm merging this.

PR #13, opened 2020-03-11
Zzy sparse tensor

zzy82518996:
This PR extends pytorch to support the sparse hammerblade backend.

The interface of part of the sparse tensor operators is extended.
Right now, you can create a sparse tensor on CPU side, and copy the data onto hammerblade device.


aten/src/ATen/native/native_functions.yaml
```diff
@@ -4517,6 +4539,7 @@
   dispatch:
     CPU: legacy::cpu::_th_nonzero
     CUDA: legacy::cuda::_th_nonzero
+    HAMMERBLADE: legacy::cuda::_th_nonzero
```
yodada: The yaml file is case sensitive, so `HAMMERBLADE` will case a building failure. Also, why we are calling `legacy::cuda::_th_nonzero` for HammerBlade?

setup_cosim_build_env.sh
```diff
@@ -30,7 +30,7 @@ echo "  hb-pytorch lives in $DIR"
 # setup cudalite runtime and pytorch kernel binary paths
 if [ -z "$BRG_BSG_BLADERUNNER_DIR" ]
 then
-  export BSG_MANYCORE_DIR="<path-to-your-cudalite-cosim-runtime>"
+  export BSG_MANYCORE_DIR="/home/zz546/software/hammerblade/bsg_bladerunner/bsg_replicant/libraries"
```
yodada, line 33: oops. don't commit your local path ;)

aten/src/ATen/native/native_functions.yaml
```diff
@@ -280,6 +280,7 @@
     HammerBlade: add
     SparseCPU: add_sparse
     SparseCUDA: add_sparse
+    SparseHammerBlade: add_sparse
```
yodada: @vb000 I think @zzy82518996 said he didn't actually implement those kernels, and these are what he intended to implement. What do you think? Should we not declare a kernel if it is not implemented yet? I kind of did the same thing ... but it was for the undergrads so they don't have to touch native_functions.yaml

vb000: I see it's a nice way to figure out how call sequence works. A dummy kernel with an error message sounds good?

vb000: Basically, anyway we could know where the error occurred. If not implementing anything would generate a reasonable error message, I guess that'd be fine as well!

yodada: My concern is that, say we define `add_sparse` here for HammerBlade, and it calls xxx_stub eventually, then the error message is terrible. It tells you _some_ stub is not implemented for HammerBlade but does not say which and where it is called.

I would advocate for not defining for HammerBlade unless we are implementing it.

hammerblade/torch/tests/test_hammerblade_sparse_backend.py
```diff
@@ -0,0 +1,98 @@
...
+    assert cpu_x.type() == 'torch.sparse.FloatTensor'
+    assert hb_x.device == hb
+    assert hb_x.type() == 'torch.hammerblade.sparse.FloatTensor'
+    assert hb_x.is_hammerblade
+    assert hb_i.type() == 'torch.hammerblade.LongTensor'
+    assert hb_v.type() == 'torch.hammerblade.FloatTensor'
+    assert torch.equal(hb_i.cpu(), cpu_i)
+    assert torch.equal(hb_v.cpu(), cpu_v)
+
+def test_move_sparsetensor_between_cup_and_hammerblade_path1():
```
yodada, line 125: `path1` is placed between `path3` and `path2`. that is a little odd.

yodada (changes requested):
I think this PR looks good in general! Only two things need to be addressed before we can hit the merge button:
 - I think we should remove the registration of HammerBlade sparse kernels in `aten/src/ATen/native/native_functions.yaml` since we don't really have them implemented
 - We need more tests. Three seems to be too few. ;)

torch/csrc/utils/tensor_types.cpp
```diff
@@ -22,6 +22,7 @@ static const char* backend_to_string(const at::Backend& backend) {
     case at::Backend::HammerBlade: return "torch.hammerblade";
     case at::Backend::SparseCPU: return "torch.sparse";
     case at::Backend::SparseCUDA: return "torch.cuda.sparse";
+    case at::Backend::SparseHammerBlade: return "torch.hammerblade.sparse";
```
vb000, line 25: Function `all_declared_types` at line 89 of this file might need modification as well...?

yodada: That is a good point! Thanks @vb000 !

yodada:
Seems you have opened a new PR. Should we close this one?

zzy82518996:
Yes, you can close this one. Thanks Lin.

----- 原始邮件 -----
发件人: "Lin Cheng" <notifications@github.com>
收件人: "cornell-brg/hb-pytorch" <hb-pytorch@noreply.github.com>
抄送: "zzy82158996" <zzy82158996@sjtu.edu.cn>, "Mention" <mention@noreply.github.com>
发送时间: 星期二, 2020年 3 月 24日 下午 12:33:10
主题: Re: [cornell-brg/hb-pytorch] Zzy sparse tensor (#13)

Seems you have opened a new PR. Should we close this one?

yodada:
Moved to #20

PR #17, opened 2020-03-19
Conv2d bias

vb000:
- Conv2d bias kernel
- A testcase for benchmarking Lenet-5 CNN on MNIST.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ before_install:
 install:
     # Install Python requirements
     - pip install --upgrade pip setuptools twine
-    - pip install numpy pyyaml mkl mkl-include setuptools cmake cffi typing sklearn tqdm pytest
+    - pip install numpy pyyaml mkl mkl-include setuptools cmake cffi typing sklearn tqdm pytest ninja torchvision
     - pip list
 
 #------------------------------------------------------------------------
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,7 +102,7 @@ jobs:
               # Enable emulation layer
               - export USE_HB_EMUL=1
               # Build and install pytorch
-              - python setup.py install
+              - python setup.py develop
               # Check torch
               - pip list
               # Run pytests
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ before_install:
 install:
     # Install Python requirements
     - pip install --upgrade pip setuptools twine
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -102,7 +102,7 @@ jobs:
               # Enable emulation layer
               - export USE_HB_EMUL=1
               # Build and install pytorch
-              - python setup.py install
+              - python setup.py develop
```
yodada, line 105: I made the same change. but let's keep it here. Otherwise I think travis will error out due to log being too long.

aten/src/ATen/native/hammerblade/Threshold.cpp
```diff
@@ -12,11 +13,10 @@ static void threshold_kernel_hb(
     Scalar threshold_scalar,
     Scalar value_scalar) {
 
-  TORCH_CHECK(false, "threshold_kernel_hb not implemented");
-    //TODO: implement the host code for threshold kernel
-    //      you may refer to add_kernel_hb in
-    //      aten/src/ATen/native/hammerblade/AddSub.cpp
-    //      as an example
+  AT_DISPATCH_FLOAT_TYPE_ONLY(iter.dtype(), "threshold_hb", [&]() {
```
yodada: This is awesome but ... we do need to revert it. Sorry ...

hammerblade/torch/kernel/kernel_threshold.cpp
```diff
@@ -13,15 +13,34 @@ extern "C" {
           bsg_tensor_t* other,
           float* threshold_scalar,
           float* value_scalar) {
-    // TODO: Convert uint32_t pointers to correct types
+    auto N = self->N; // number of elements
```
yodada: Same here ... :(

yodada (commented):
Hmmm. I think your stuff does need ReLU to work. I'm not sure what is correct thing to do here.
@cbatten thoughts? Should we delay this PR until we have Angela's threshold kernel?

vb000:
That's okay... I can remove threshold from this pr but keep in the other branch I'm working on. I can rebase it with Angela's kernel when available.

yodada (approved):
Looks good to me!

cbatten:
Is this just blocked on @AngelaZou2000 's ReLU kernel? I think Angela has that ready to go right?

vb000:
Right, I think branches `az292_relu` and `kz73_addmm` have implementations that can be merged in. It'd simplify further development, if they are merged...

PR #18, opened 2020-03-19
Element-wise-for + Offloading refactor + Other improvements

yodada:
Included in this PR:
 - For kernels that use TensorIterator on the host, we added helper class / functions `BRGIteratorTensor`, `brg_tile_element_wise_for` and `brg_tile_for` to make writing kernels easier. Note that they will be renamed `HB_*` later. See `kernel_add.cpp` for an example.
 - Offloading APIs are refactor to make writing the host code much easier.
 - Copy: now we have the *correct* implementation of copying between cpu/hb and hb/hb.
 - A few new kernels are working, such as `clone`, `expand` and `contiguous`
 - Dropout is fully functional

Be very careful with nullary operations. For example `hammerblade/torch/kernel/kernel_fill.cpp`
Even though a nullary op may not take any parameter, we need this `float a` to figure out the type.
```
brg_tile_element_wise_for(t0_p, [&](float a = 0.0f)
```


.travis.yml
```diff
@@ -102,9 +102,9 @@ jobs:
...
               # Build and install pytorch
-              - python setup.py install
+              - python setup.py develop
               # Check torch
               - pip list
               # Run pytests
-              - cd hammerblade/torch/tests
-              - python -m pytest -v .
+              - cd hammerblade/torch/
+              - python pytest_runner.py
```
vb000, line 110: I imagine to make sure cosim would start running without issues?

aten/src/ATen/native/hammerblade/Copy.cpp
```diff
@@ -4,16 +4,32 @@
...
+
+  bool same_dtype = iter.dtype(0) == iter.dtype(1);
+  if (same_dtype && iter.is_contiguous()) {
+    // Contiguous same-dtype copies can always use memcpy
+    return false;
+  }
+  return true;
+}
 
 static void copy_hb_to_hb(TensorIterator& iter, bool non_blocking) {
```
vb000, line 31: Why not overload the function `offload_memcpy` to call `tensorlib_copy_hb_to_hb` when passed iterators. Or, totally remove `offload_memcpy` and keep this? Two functions doing the same thing make API little confusing...

aten/src/ATen/native/hammerblade/Distributions.cpp
```diff
@@ -0,0 +1,40 @@
...
+  AT_DISPATCH_FLOAT_TYPE_ONLY(self.scalar_type(), "bernoulli_scalar_hb_", [&]() {
+    HammerBladeGenerator* generator = get_generator_or_default<HammerBladeGenerator>(gen,
+                                      hammerblade::detail::getDefaultHammerBladeGenerator());
+    auto p_scalar = Scalar(p);
+    auto seed_tensor = hammerblade_common_seed_to_tensor(generator->random());
+
+    hb_offload_kernel(self, seed_tensor, p_scalar, "tensorlib_bernoulli_scalar_");
+  });
+  return self;
+}
```
vb000, line 38: Is it possible to add small comment about what this function does? This feels like an important routine that's going to be useful in more randomization based algorithms like dropout.

aten/src/ATen/native/hammerblade/Empty.cpp
```diff
@@ -58,4 +43,10 @@ Tensor empty_hb(IntArrayRef size, const TensorOptions& options, c10::optional<c1
   return tensor;
 }
 
+Tensor empty_strided_hb(IntArrayRef size, IntArrayRef stride, const TensorOptions& options) {
+  auto t = at::native::empty_hb({0}, options);
+  at::native::resize_impl_hb_(t.unsafeGetTensorImpl(), size, stride);
+  return t;
+}
```
vb000, line 50: Just curious... is this to create empty tensors with strides of our choice?

aten/src/ATen/native/hammerblade/Offload.h
```diff
@@ -1,102 +1,138 @@
...
+                              Scalar s0, const char* kernel) {
+  std::vector<Tensor> args;
+  args.push_back(t0);
+  args.push_back(t1);
+  args.push_back(t2);
+  args.push_back(t3);
+  std::vector<Scalar> scalars;
+  scalars.push_back(s0);
+  offload_tensor_scalar_impl(args, scalars, kernel);
+}
```
vb000, line 130: This looks great! 

I spent a day or so looking for variadic templates to make this more elegant... this needs "compile-time looping". I found c++11 doesn't have much support for that.

aten/src/ATen/native/hammerblade/Offload.h
```diff
@@ -1,102 +1,138 @@
...
+
+//==============================================
+// conv offloader
+//==============================================
+
 void offload_convolution_forward(Tensor& output, const Tensor& input,
     const Tensor& weight, IntArrayRef padding, IntArrayRef stride,
     IntArrayRef dilation, int64_t groups);
```
vb000, line 138: I'll move this to `Convolution.cpp` in my next pr.

hammerblade/torch/kernel/brg_element_for.hpp
```diff
@@ -0,0 +1,356 @@
...
+//
+// We would like the synatx to be like this:
+// __attribute__ ((noinline))  int tensorlib_add(bsg_tensor_t* c_p,
+//                                               bsg_tensor_t* a_p,
+//                                               bsg_tensor_t* b_p,
+//                                               float* alpha) {
+//    brg_elementwise_for(t0_p, t1_p, t2_p,
+//        [&](float a, float b) {
+//          return a + alpha * b;
+//        });
```
vb000, line 127: Thought we decided to hold-off lambda functions for a while... jk, I love this. This helps us decouple optimizing parallel primitives vs optimizing actual computation.

vb000 (approved):

aten/src/ATen/native/hammerblade/Copy.cpp
yodada: Yeah I can totally see what you are saying. I like the overloading idea. Making changes now ... but it will be in the merge branch, not here ;)

aten/src/ATen/native/hammerblade/Distributions.cpp
yodada: Will do!

aten/src/ATen/native/hammerblade/Empty.cpp
yodada: Good question. I have to admit that I'm not 100% sure what it does, but I think you are right

aten/src/ATen/native/hammerblade/Offload.h
yodada: yeah that would super hard I guess

aten/src/ATen/native/hammerblade/Offload.h
yodada: Perfect!

hammerblade/torch/kernel/brg_element_for.hpp
yodada: oh well you know ... things happen ...

So the cool part here is that we figure out the data type by looking at the lambda function ;)

yodada:
Switching to lc873-element-wise-merge branch

PR #19, opened 2020-03-20
Element-wise-for + Offloading refactor + Other improvements - conflicts-resolved

yodada:
Included in this PR:
 - For kernels that use TensorIterator on the host, we added helper class / functions `BRGIteratorTensor`, `brg_tile_element_wise_for` and `brg_tile_for` to make writing kernels easier. Note that they will be renamed `HB_*` later. See `kernel_add.cpp` for an example.
 - Offloading APIs are refactor to make writing the host code much easier.
 - Copy: now we have the *correct* implementation of copying between cpu/hb and hb/hb.
 - A few new kernels are working, such as `clone`, `expand` and `contiguous`
 - Dropout is fully functional

Be very careful with nullary operations. For example `hammerblade/torch/kernel/kernel_fill.cpp`
Even though a nullary op may not take any parameter, we need this `float a` to figure out the type.
```
brg_tile_element_wise_for(t0_p, [&](float a = 0.0f)
```


vb000 (approved):
Thanks for the updates! Asserts are great!

PR #21, opened 2020-03-23
Maxpool and softmax

vb000:
- `MaxPool2d` and `LogSoftmax` kernels.
- Moved convolution offloading to `Convolution.cpp`
- Fixed dispatch warnings.
- Filter out `-nostdlib` flag when compiling kernels for cosim -- to use C math library.

### DIFF
--- a/aten/src/ATen/Dispatch.h
+++ b/aten/src/ATen/Dispatch.h
@@ -8,7 +8,7 @@
 
 #define AT_PRIVATE_CASE_TYPE(enum_type, type, ...) \
   case enum_type: {                                \
-    using scalar_t = type;                         \
+    using scalar_t C10_UNUSED = type;              \
     return __VA_ARGS__();                          \
   }
 
--- a/aten/src/ATen/native/native_functions.yaml
+++ b/aten/src/ATen/native/native_functions.yaml
@@ -5896,6 +5896,7 @@
   dispatch:
     CPU: max_pool2d_with_indices_cpu
     CUDA: max_pool2d_with_indices_cuda
+    HammerBlade: max_pool2d_with_indices_hb
 
 - func: max_pool2d_with_indices_backward.grad_input(Tensor grad_output, Tensor self, int[2] kernel_size, int[2] stride, int[2] padding, int[2] dilation, bool ceil_mode, Tensor indices, *, Tensor(a!) grad_input) -> Tensor(a!)
   python_module: nn
--- a/aten/src/ATen/native/native_functions.yaml
+++ b/aten/src/ATen/native/native_functions.yaml
@@ -1586,6 +1586,7 @@
   dispatch:
     CPU: log_softmax_cpu
     CUDA: log_softmax_cuda
+    HammerBlade: log_softmax_hb
 
 - func: _log_softmax_backward_data(Tensor grad_output, Tensor output, int dim, Tensor self) -> Tensor
   use_c10_dispatcher: full
--- a/hammerblade/torch/kernel/kernel_pooling.cpp
+++ b/hammerblade/torch/kernel/kernel_pooling.cpp
@@ -21,8 +21,6 @@ extern "C" {
     // Start profiling
     bsg_cuda_print_stat_kernel_start();
 
-    exit(1);
-
     // End profiling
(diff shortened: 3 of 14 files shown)


vb000:
All pass on cosim.
```
============ 88 passed, 17 skipped, 2 xfailed in 697.64s (0:11:37) =============
BSG REGRESSION TEST PASSED
BSG COSIM PASS: Test passed!
$finish called from file "/mnt/users/ssd1/no_backup/bandhav/bsg_bladerunner/bsg_replicant/testbenches/cosim_wrapper.sv", line 57.
$finish at simulation time         191369340000
           V C S   S i m u l a t i o n   R e p o r t 
Time: 191369340000 ps
CPU Time:    701.250 seconds;       Data structure size:   3.6Mb
Mon Mar 23 15:08:15 2020
```

aten/src/ATen/native/native_functions.yaml
```diff
@@ -1586,6 +1586,7 @@
   dispatch:
     CPU: log_softmax_cpu
     CUDA: log_softmax_cuda
+    HammerBlade: log_softmax_hb
```
yodada, line 1589: Exactly what I need! Sweet!

aten/src/ATen/Dispatch.h
```diff
@@ -8,7 +8,7 @@
 
 #define AT_PRIVATE_CASE_TYPE(enum_type, type, ...) \
   case enum_type: {                                \
-    using scalar_t = type;                         \
+    using scalar_t C10_UNUSED = type;              \
```
yodada, line 11: What is the usage of this? I think I'm missing sth

hammerblade/torch/kernel/kernel_pooling.cpp
```diff
@@ -0,0 +1,85 @@
...
+          bsg_tensor_t* output,
+          bsg_tensor_t* input,
+          bsg_tensor_t* indices,
+          int* kH, int* kW,
+          int* sH, int* sW,
+          int* padH, int* padW,
+          int* dilationH, int* dilationW) {
+    auto y = BSGTensor<float>(output);
+    auto x = BSGTensor<float>(input);
+    auto ind = BSGTensor<long long>(indices);
```
yodada: so we can do 64bit arithmetic on HB?

aten/src/ATen/native/native_functions.yaml
vb000: Yep, almost every modern classification network uses this as the last layer.

aten/src/ATen/Dispatch.h
vb000: Dispatch macro sets `scalar_t` with `using scalar_t C10_UNUSED = type; `. This may or may not be used in the offloading function. This results in compiler warning about unused typedef. Adding `C10_UNUSED` adds a compile-time attribute to ignore that warning.

aten/src/ATen/native/native_functions.yaml
yodada: Have you looked at any loss function yet? I'm trying to get `CrossEntropyLoss` to work. It somehow uses this logSoftMax kernel

hammerblade/torch/kernel/kernel_pooling.cpp
vb000: `int64_t` to `int32_t` could be a destructive cast. So I was hesitant to go ahead and cast it. But in the hindsight, that seems like the right thing to do.

aten/src/ATen/Dispatch.h
yodada: Ok! Understand!

aten/src/ATen/native/native_functions.yaml
vb000: No, haven't looked at loss functions yet. Those are definitely needed for training!

hammerblade/torch/kernel/kernel_pooling.cpp
yodada: Maybe make this change and otherwise I think this PR is in good shape!

yodada (approved):
Looks good! Feel free to merge after getting rid of that `long long`! ;)

PR #23, opened 2020-03-26
LogSoftMax backward

yodada:
Implemented LogSoftMax backward kernel.
With this PR, torch.nn.NLLLoss and torch.nn.CrossEntropyLoss are functional.


yodada:
Running cosim now ... there is a floating point issue in `test_div_with_scalar_hypothesis`. Hypothesis tends to generates a tiny scalar (10^-36) and that test fails.

hammerblade/torch/kernel/kernel_softmax_backward.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+  __attribute__ ((noinline))  int tensorlib_log_softmax_backward(
+          bsg_tensor_t* grad_input_p,
+          bsg_tensor_t* grad_p,
+          bsg_tensor_t* output_p,
+          int32_t*      dim_p) {
+
+    BSGTensor<float> gI(grad_input_p);
+    BSGTensor<float> grad(grad_p);
+    BSGTensor<float> output(output_p);
```
vb000: Guess these are unused?

hammerblade/torch/tests/test_log_softmax_backward.py
```diff
@@ -0,0 +1,45 @@
...
+                         requires_grad=True)
+    input_h = torch.tensor([[-3.5164, -1.9673, -4.7514, -0.2075, -4.6912],
+                           [-2.9321, -1.5085, -0.8699, -1.4016, -2.8089],
+                           [-2.6460, -1.9800, -1.4818, -1.8358, -0.9057]],
+                           requires_grad=True)
+    target = torch.tensor([1, 0, 4])
+    output = loss(m(input), target)
+    logsoftmax = m(input_h.hammerblade())
+    assert logsoftmax.device == torch.device("hammerblade")
+    output_h = loss(logsoftmax.cpu(), target)
```
vb000: Is it required to do backward on `loss` and not directly on `LogSfotmax` to test `log_softmax_backward`? I'd imagine later would be a faster test...?

vb000 (approved):
Thanks for smalls prs! Easy to follow and sync on chnages ;)

hammerblade/torch/kernel/kernel_softmax_backward.cpp
yodada: I think this is true for `gI` and `output`

hammerblade/torch/tests/test_log_softmax_backward.py
yodada: Yes, technically we can. I was just lazy and reused old tests (with different parts on cpu) XD

yodada: Updated to directly test logsoftmax_backward! Maybe take a look and see if there is anything wrong?

vb000: Looks great! Any advantage with `np.random` over `torch.rand`? (I'll try to use that in my tests if so...)

yodada: I think the reason is just I want to create two identical tensors (without clone etc). I don't think one is better than the other ;)

vb000: Got it.

PR #28, opened 2020-03-29
Con2d & MaxPool2d backward

vb000:
- Conv2d backward operator.
- MaxPool2d backward operator.
- Added `bsg_assert_msg` to print error messages with assert.
- Place read-only data section in dram, not dmem.

### DIFF
--- a/hammerblade/torch/hb_pytorch_link.ld
+++ b/hammerblade/torch/hb_pytorch_link.ld
@@ -1,0 +1,184 @@
+/******************************/
+/*                            */
+/*   Manycore linker script   */
+/*                            */
+/******************************/
+
+/* This link script places data sections in dmem
+ * by default. Text sections are placed in dram.
+ *
+ * Data in sections ".rodata.*" and ".dram"
+ * is placed in the DRAM.
+ */
+
+
+/* Constants */
+PROVIDE(_dmem_lma_start   = 0x00001000);
+PROVIDE(_dram_t_lma_start = 0x80000000);
+PROVIDE(_dram_d_lma_start = 0x81000000);
+
+
+
+/*=============================
+ * Memory region descriptions
+ *=============================*/
+
+MEMORY
+{
+  /* Local memory VMA */
+  DMEM_VMA (rw) : ORIGIN = 0x1000, LENGTH = 4K
+
+  /* DRAM text VMA */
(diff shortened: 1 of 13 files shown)


hammerblade/torch/hb_pytorch_link.ld
```diff
@@ -0,0 +1,184 @@
+/******************************/
```
yodada, line 1: This looks very important. Could you maybe lecture us what is difference between dram and dmem?

hammerblade/torch/kernel/bsg_assert.hpp
```diff
@@ -5,8 +5,13 @@
 #ifndef _BSG_ASSERT_HPP
 #define _BSG_ASSERT_HPP
 
-#define bsg_assert(cond) if (!(cond)) {                                          \
-    bsg_printf("assert failed at %s:%d\n", __FILE__, __LINE__);                \
+#define bsg_assert(cond) if (!(cond)) {                       \
+    bsg_printf("assert failed at %s:%d", __FILE__, __LINE__); \
+    bsg_fail();}
+
+#define bsg_assert_msg(cond, fmt, ...) if (!(cond)) {          \
```
yodada, line 12: sweet!

hammerblade/torch/tests/hbutils.py
```diff
@@ -0,0 +1,6 @@
+def init_hb_tensor(input_t):
+    """
+    Returns a HB tensor that's a leaf node of the computation graph.
+    """
+    return input_t.hammerblade().clone().detach().requires_grad_(
```
yodada, line 5: This is very smart! apparently I don't know much about using pytorch and ML in general. ;)

yodada (approved):
Looks good to me!

hammerblade/torch/hb_pytorch_link.ld
vb000: `dmem` is the scratch pad and `dram` is the main memory of HB. Debug messages like printing and constants are stored in `rodata` section, which by default goes into `dmem` in Manycore's link script. We would want those to go to `dram` as scratchpad is a scarce resource.

yodada: ahh I see.

PR #29, opened 2020-03-30
single tile addmm with blocking

kzheng9:
Single tile addmm with blocking implemented. However, may have some floating point precision issues when handling large matrices. Can resolve this by lowering tolerance in the test code: `torch.allclose(x, y, atol=1e-07, rtol=1e-04)`

Also included in this PR: (@yodada 's commits)
 - fixed a bug in bernoulli_scalar_ in which the probability we get is actually 1-p
 - implemented `torch.tensor.item`, in which we read the sole element from a 1-element tensor. 
 - Host code template for `torch.sum`. Note: I (@yodada) have a "working" sum kernel in my local branch, but it only "works". To get performant reduction ops, we need to change a lot of things. @vb000 we need to sit down and discuss this.


yodada:
@kzheng9  It looks like you have a few python style issue in your test files.

yodada:
Looks like this PR contains a few commits from my mnist branch. That is ok, but we'll need to have @vb000 's review. Also, you might want to revert commit `6fc57aa` because it has my ReLU implementation

aten/src/ATen/native/hammerblade/Addmm.cpp
```diff
@@ -12,13 +13,24 @@ Tensor addmm_hb(
...
-  return self;
+  if ( (self.scalar_type() != ScalarType::Float)
+    || (mat1.scalar_type() != ScalarType::Float)
+    || (mat2.scalar_type() != ScalarType::Float) ) {
+    AT_ERROR("HammerBlade addmm is implemented for Float only");
+  }
 
+  using scalar_t = float;
+
+  TORCH_CHECK(mat1.size(1) == mat2.size(0), "Argument #3: Expected dim 0 size ", mat1.size(1), ", got ", mat2.size(0));
```
vb000, line 26: This is a nice reference for argument checks required for addmm, I think a couple more checks also makes sense:
https://github.com/cornell-brg/hb-pytorch/blob/d5358b65dd034b15d5a8d593436966c11b0361be/aten/src/TH/generic/THTensorMath.cpp#L130-L147

kzheng9: Actually the `expand_size` function takes care of broadcasting/mismatched matrices here: `std::tie(b_self) = expand_size(self, {mat1.size(0), mat2.size(1)}, "addmm_out");` Which is why I did not include those checks. Do you think they are still necessary?

yodada: I think it might be still useful to check if mat1 and mat2 are 2D tensors (matrices)?

hammerblade/torch/tests/test_addmm.py
```diff
@@ -3,9 +3,9 @@
 03/09/2020 Lin Cheng (lc873@cornell.edu)
 """
 import torch
-import pytest
+# import pytest
```
yodada: I would suggest removing these commented code.

hammerblade/torch/tests/test_mm.py
```diff
@@ -25,13 +25,24 @@ def test_torch_mm_2():
...
 @pytest.mark.xfail
+# @pytest.mark.skip(reason="not yet implemented")
 def test_torch_mm_mismatching_shape_F():
     mat1 = torch.randn(2, 2)
     mat2 = torch.randn(3, 3)
     mat1_h = mat1.hammerblade()
     mat2_h = mat2.hammerblade()
     out_h = torch.mm(mat1_h, mat2_h)
+
+# @pytest.mark.skip(reason="not yet implemented")
```
yodada: same here

aten/src/ATen/native/hammerblade/Addmm.cpp
kzheng9: Added!

hammerblade/torch/tests/test_mm.py
kzheng9: Removed.

hammerblade/torch/tests/test_addmm.py
kzheng9: Removed.

hammerblade/torch/kernel/kernel_addmm.cpp
```diff
@@ -1,29 +1,183 @@
...
     bsg_cuda_print_stat_kernel_start();
+
     // TODO: Implement addmm
+
+    // v2: single tile, use blocking
+    int r1 = mat1.dim(0);
+    int c1 = mat1.dim(1);
+    int r2 = mat2.dim(0);
+    int c2 = mat2.dim(1);
+    assert(c1 == r2);
```
vb000: `assert` calls `exit` which has undefined behavior on the device. Since this is device code, we'd want to exit using device mechanisms. HammerBlade's exit mechanism is to call `bsg_fail()`. Lin implemented an assertion wrapper around it called `bsg_assert`. You could use that here...

hammerblade/torch/kernel/kernel_addmm.cpp
```diff
@@ -1,29 +1,183 @@
...
+
+    bsg_printf("r1 %d c1 %d r2 %d c2 %d\n", r1, c1, r2, c2);
+    bsg_printf("BLOCK_DIM %d\n", BLOCK_DIM);
```
vb000: You might want to remove these printfs...

hammerblade/torch/kernel/kernel_addmm.cpp
```diff
@@ -1,29 +1,183 @@
...
+
+    // calculate number of row and col blocks in each matrix
+    int m1_num_blk_per_row = std::ceil((float)r1/(float)BLOCK_DIM); // how many blocks in m1 per row
+    int m1_num_blk_per_col = std::ceil((float)c1/(float)BLOCK_DIM); // how many blocks in m1 per col
+    int m2_num_blk_per_row = std::ceil((float)r2/(float)BLOCK_DIM); // how many blocks in m2 per row
+    int m2_num_blk_per_col = std::ceil((float)c2/(float)BLOCK_DIM); // how many blocks in m2 per col
```
vb000: In terms of the optimizations, we'd rather prefer to have parallel naive implementations than high-performant single tile implementations. It could be more beneficial to put efforts in parallelization of your implementation first...

Moreover, classical matrix multiplication optimization might not work well for HB because of it's unique memory system.

cbatten: Hmmm ... blocking is critical for both the single tile and the parallel implementation. Also keep in mind that we are just getting started. And @kzheng9 and I agreed that we would start by focusing on a detailed data-drive optimization using blocking with a single tile and make sure we really understand that before moving on to parallelizing it (which should be relatively straight-forward once we have blocking working) ... and also keep in mind that we are well aware of the "unique memory system" of HB :)

cbatten: @kzheng9 avoid using floating point arithmetic to do this kind of block range calculation ... see if you can use integer arithmetic at all costs otherwise it just adds overhead

yodada:
I kind of want to tease out my commits and put them in another PR ... I can cherry-pick your commits (and my commits merging the addmm) for you.  Ideas? @kzheng9

hammerblade/torch/kernel/kernel_addmm.cpp
kzheng9: Will do!

hammerblade/torch/kernel/kernel_addmm.cpp
kzheng9: Whoops, sorry about that. Removed..

kzheng9:
@yodada The worry I have about cherry-picking my commits only is that it might cause it to not be able to pass any tests, since some of the tests require the fix to the non-contiguous memory access (like we dicussed about) to work. Perhaps we could cherry-pick both my commits and that fix?

hammerblade/torch/kernel/kernel_addmm.cpp
kzheng9: Got it! I have modified my code so it only does integer arithmetic.

yodada:
Take a look at this branch. I cherry-picked all your commits (I hope?) 
https://github.com/cornell-brg/hb-pytorch/tree/merge_addmm

kzheng9:
Looks good to me! I've opened a PR here: https://github.com/cornell-brg/hb-pytorch/pull/32. 

Closing this one.

PR #30, opened 2020-03-30
parallel relu kernel

AngelaZou2000:
Implemented relu kernel with element_wise_for so can run on multiple tiles. Also passed hypothesis testing.

### DIFF
--- a/aten/src/ATen/native/hammerblade/Offload.cpp
+++ b/aten/src/ATen/native/hammerblade/Offload.cpp
@@ -171,7 +171,6 @@ void offload_tensor_scalar_impl(std::vector<Tensor> tensors,
 
 }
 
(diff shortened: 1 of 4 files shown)


aten/src/ATen/native/hammerblade/Offload.cpp
```diff
@@ -48,6 +171,222 @@ void offload_tensor_scalar_impl(std::vector<Tensor> tensors,
 
 }
 
+<<<<<<< HEAD
```
yodada: Ooops. you have unresolved merge conflicts here

AngelaZou2000: yep I deleted that!

PR #32, opened 2020-03-31
Merge addmm

kzheng9:
Merging cherry-picked addmm. See https://github.com/cornell-brg/hb-pytorch/pull/29

PR #31, opened 2020-03-30
relu kernel

AngelaZou2000:
Parallel relu kernel implemented.

PR #34, opened 2020-04-01
Dropout fix + torch.tensor.item()

yodada:
Included in this PR:
 - fixed a bug in `bernoulli_scalar_`. The probability we received in this kernel is *(1-p)*
 - implemented `torch.tensor.item()` --> kernel `_local_scalar_dense` 
 - force pytorch to run with a single thread. disable OMP alone is not sufficient to prevent parallelization
 - Removed fake mean implementation --- it shouldn't be there in the first place. my bad

With this PR + #31 + #32 + #36 , one should be able to train a multilayer perceptron model for MNIST


setup_cosim_build_env.sh
```diff
@@ -17,6 +17,7 @@ export USE_DISTRIBUTED=0
 export USE_OPENMP=0
 export ATEN_THREADING=NATIVE
 export CFLAGS='-fuse-ld=gold'
+export OMP_NUM_THREADS=1
```
vb000, line 20: USE_OPENMP=0?

vb000 (approved):
Looks good!

setup_cosim_build_env.sh
yodada: Somehow pytorch also uses this `OMP_NUM_THREADS` to control its pthread pool.
https://github.com/pytorch/pytorch/issues/975#issuecomment-396392836
just playing safe. because when i was training on CPU, I did see a lot of worker threads

PR #36, opened 2020-04-02
Sum and binary reduction

yodada:
In this PR:
 - we can now do reduction on iterators that have at most 3D shapes.
 - implemented `torch.sum`

To use `binary_reduction`, we need to define 2 lambda functions:
 - `reduce` takes the partial result and an input element. it defines how to reduce an input element.
    For `sum`, reduce is `partial_result += input`
 - `project` takes in the reduced result and performs postprocessing. For `sum`, project returns the identity. But for `mean`, it returns `result /  num_elements_per_output`

This is a joint PR with @vb000 , so it would be great if @zzy82518996 could also take a look and review this one.

### DIFF
--- a/aten/src/ATen/native/ReduceOps.cpp
+++ b/aten/src/ATen/native/ReduceOps.cpp
@@ -260,6 +260,7 @@ Tensor& sum_out(Tensor& result, const Tensor& self, IntArrayRef dim,
   if (iter.numel() == 0) {
     result.zero_();
   } else {
+    result.zero_();
     sum_stub(iter.device_type(), iter);
   }
   return result;
--- a/aten/src/ATen/native/hammerblade/OffloadIter.h
+++ b/aten/src/ATen/native/hammerblade/OffloadIter.h
@@ -18,6 +18,71 @@ namespace native {
 void offload_iterator_op_impl(TensorIterator& iter, std::vector<eva_t> device_scalars,
     const char* kernel, uint32_t ntensors);
 
+//=======================================================================
+// Offloading operations that use reduction TensorIterator
+//=======================================================================
+
+template<typename scalar_type>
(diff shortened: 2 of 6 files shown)


zzy82518996:
I have a question, I saw there is also sum operation for sparse tensor. May I ask in what scenario do we need this operator ?

aten/src/ATen/native/ReduceOps.cpp
```diff
@@ -260,6 +260,7 @@ Tensor& sum_out(Tensor& result, const Tensor& self, IntArrayRef dim,
   if (iter.numel() == 0) {
     result.zero_();
   } else {
+    result.zero_();
```
vb000, line 263: Just one question, don't we want to accelerate this on device?

vb000: Oh, it might be hb to hb copy, am I right?

yodada: It will call kernel `fill_`. The only reason we need this, for now, is because of the trivial case.
If the input tensor is not contiguous, the trivial kernel will be called multiple times. In the next PR I hope to fix & parallelize the trivial case.

yodada:
@zzy82518996 we are implementing this one since it is used when running the backwards pass:
https://github.com/cornell-brg/hb-pytorch/blob/3a7ef13c92597d72d507be01ab2ebaf0737f2bb2/torch/csrc/autograd/engine.cpp#L452

aten/src/ATen/native/ReduceOps.cpp
vb000: Oh okay. Calling `fill_` also does the job, I guess. Thanks for the explaining why it's needed, I couldn't totally understand it before.

vb000 (approved):

PR #38, opened 2020-04-03
CheckLayer: a debug layer to cross-check HB intermediate results

vb000:
`CheckLayer` is a debug layer (a sub-class of `torch.nn.Module`) that is intended to simplify root causing the misbehaving HB layer. It's used like this:

https://github.com/cornell-brg/hb-pytorch/blob/c3c0a5745af45b39e34aea5a57a4ecc5d61f66ec/hammerblade/torch/tests/hbutils.py#L16-L58

This debug layer goes slightly against the grain with PyTorch autograd. So, suggestions to make it better or more useful are welcome...


yodada (approved):
Nice! I think it is a very useful tool.

Issue #40, opened 2020-04-03
Fix build on macOS


sampsyo:
I naively thought that building hb-pytorch on my Mac would be problem-free for a fun on-laptop development experience, but I've run into a few miscellaneous issues. I'll keep track of them here and open individual pull requests with fixes. I also have a few experimental changes in the [mac-build](https://github.com/cornell-brg/hb-pytorch/tree/mac-build) branch, but that's not meant to be merged.

Here's what I've found so far:

- [x] The `-fuse-ld=gold` [flag](https://github.com/cornell-brg/hb-pytorch/blob/cc9804d9d072645d95f505b93d577c622d8ca2c7/setup_emul_build_env.sh#L18) to the C compiler needs to be deleted or switched to use plain old ld as the linker. (Any particular reason gold is required in the first place?)
- [x] The `features.h` header does not exist. It should be harmless to replace [this include](https://github.com/cornell-brg/hb-pytorch/blob/cc9804d9d072645d95f505b93d577c622d8ca2c7/c10/hammerblade/emul/bsg_manycore_features.h#L31) with some one-off hacks.
- [ ] The `hb_mc_idx_t` type is [typedef'd](https://github.com/cornell-brg/hb-pytorch/blob/cc9804d9d072645d95f505b93d577c622d8ca2c7/c10/hammerblade/emul/bsg_manycore_coordinate.h#L53) to be 32 bits, but it is used in several printfs as an 8-bit char. This triggers an error (I think it's a clang warning, and something is using -Werror to turn it into an error?).

vb000:
`gold` is just to make runs faster... reportedly it's faster than `ld`. Removing that wouldn't break any functionality. 

I think `-fuse-ld` flag is only available in later versions of gcc.

sampsyo:
Thanks! Maybe I'll add a check to see whether gold is installed.

sampsyo:
OK, that's all the PRs for now—with those, I can build on my Mac. There are still some test failures that we can investigate.

sampsyo:
Yay!! Things are now working for me (except for a couple of tests, which seem like the "hard" ones).

<img width="682" alt="Screen Shot 2020-04-14 at 8 47 28 AM" src="https://user-images.githubusercontent.com/188033/79226545-957ed700-7e2c-11ea-9914-493e02887fbc.png">

yodada:
I guess the dot_hypothesis one is another floating point problem? :(

sampsyo:
[...]

PR #41, opened 2020-04-03
Setup scripts: make -fuse-ld flag "optional"

sampsyo:
This is part of #40 (my effort to build hb-pytorch on macOS). The setup scripts include a CFLAGS setting to use [gold](https://en.wikipedia.org/wiki/Gold_(linker)) as the linker. In https://github.com/cornell-brg/hb-pytorch/issues/40#issuecomment-608623454, @vb000 clarified that this is just to make linking faster. These changes enable that flag only if the system actually has gold installed.


vb000 (approved):
Thanks for adding the check @sampsyo. We had only been working with devtoolset-8, so overlooked this...

PR #42, opened 2020-04-03
Remove dependency on glibc-specific header, features.h

sampsyo:
This header comes from the original manycore libraries:
https://github.com/bespoke-silicon-group/bsg_replicant/blob/master/libraries/bsg_manycore_features.h#L31

The header's sole purpose is to include `features.h` (a glibc-specific header) and check some compiler versioning stuff. Its inclusion prevents compilation on macOS (cf #40). AFAIK, I don't think we need this version stuff for emulation. So this PR just turns the header into a no-op stub. (I considered removing it altogether, but it seemed nice to avoid modifying all the places where it's `#include`d so it's possible to copy over new versions from the Replicant repository when they're updated without modification.)

### DIFF
--- a/c10/hammerblade/emul/bsg_manycore_features.h
+++ b/c10/hammerblade/emul/bsg_manycore_features.h
@@ -26,23 +26,13 @@
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #ifndef BSG_MANYCORE_FEATURES_H
-#define BSG_MANYCORE_FEATURES_H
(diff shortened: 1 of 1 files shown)


yodada (approved):
Great! Actually I never figured out what exactly is this _XOPEN_SOURC macro :(

sampsyo:
Indeed. TBH, I'm not sure this stuff is even necessary for the "real" HB libraries… it [can enable stuff like strdup](https://stackoverflow.com/a/5724485/39182), but it's not clear if that's actually necessary anywhere in the codebase.

PR #43, opened 2020-04-03
Add missing `#include <string>`

sampsyo:
Part of #40. The `kernel_trampoline.h` header uses `std::string` but does not include the appropriate header. (Clang emits the obvious error; I am not sure why gcc does not complain.)


yodada (approved):
Thanks for the fix Adrian!

PR #44, opened 2020-04-03
Remove illegal & unnecessary variable-length array initializer in addmm kernel

### LINKED ISSUE #35: Addmm failing for some dimensions
## 🐛 Bug

Apparently, `addmm_hb` kernel is failing for dimensions greater than 10, possibly due to index mis-calculation with `BLOACK_SIZE = 10`.

## To Reproduce

1. Run the following testcase:

```
    def test_torch_addmm_6():
        M = torch.rand(1, 11)
        M_h = M.hammerblade()
        mat1 = torch.rand(1, 1)
        mat1_h = mat1.hammerblade()
        mat2 = torch.rand(1, 11)
        mat2_h = mat2.hammerblade()
        out = torch.addmm(M, mat1, mat2)
        out_h = torch.addmm(M_h, mat1_h, mat2_h)
        assert out_h.device == torch.device("hammerblade")
        assert torch.allclose(out_h.cpu(), out)
```

2. You'd observe following error. Note that only 11th element is mis-computed.
```
>       assert torch.allclose(out_h.cpu(), out)
E       AssertionError: assert False
E        +  where False = <built-in method allclose of type object at 0x7fbbb836e180>(tensor([[0.8388, 0.9638, 0.9824, 0.9500, 0.9602, 0.7416, 0.2890, 0.8992, 0.1118,\n         1.1431, 0.9941]]), tensor([[0.8388, 0.9638, 0.9824, 0.9500, 0.9602, 0.7416, 0.2890, 0.8992, 0.1118,\n         1.1431, 1.1324]]))
E        +    where <built-in method allclose of type object at 0x7fbbb836e180> = torch.allclose
E        +    and   tensor([[0.8388, 0.9638, 0.9824, 0.9500, 0.9602, 0.7416, 0.2890, 0.8992, 0.1118,\n         1.1431, 0.9941]]) = <built-in method cpu of Tensor object at 0x7fbbaf037318>()
E        +      where <built-in method cpu of Tensor object at 0x7fbbaf037318> = tensor([[0.8388, 0.9638, 0.9824, 0.9500, 0.9602, 0.7416, 0.2890, 0.8992, 0.1118,\n         1.1431, 0.9941]], device='hammerblade').cpu
```

sampsyo:
Under Clang, variable-length arrays in C++ are not allowed to have initializers. (N.B. even in C99, which standardizes VLAs, [this kind of initialization is illegal](https://stackoverflow.com/a/52365995/39182).) Clang emits this error:

```
../hammerblade/torch/kernel/kernel_addmm.cpp:62:33: error: variable-sized object may not be initialized
                float sp_result[res_dim_y][res_dim_x] = {};
```

This PR just removes the initializer, which I *think* should be fine because this output array is "overwritten" anyway?

(Not terribly relevant here, but C99 VLAs in C++ are a GNU extension anyway and have [some pitfalls](https://stackoverflow.com/a/21519062/39182). I know they are the most convenient way to allocate space in the manycore's scratchpads, but in an ideal world, we'd be `malloc`ing these on the heap somehow—if it were possible to `malloc` in the scratchpad. (Maybe it already is?))

Part of #40.

### DIFF
--- a/hammerblade/torch/kernel/kernel_addmm.cpp
+++ b/hammerblade/torch/kernel/kernel_addmm.cpp
@@ -59,7 +59,7 @@ extern "C" {
                 int res_dim_x = rc == m2_num_blk_per_col - 1 ? m2_last_blk_dim_x : BLOCK_DIM;
 
                 // initialize scratchpad result
-                float sp_result[res_dim_y][res_dim_x] = {};
+                float sp_result[res_dim_y][res_dim_x];
                 for (int i = 0; i < res_dim_y; i++) {
                     for (int j = 0; j < res_dim_x; j++) {
                     }
--- a/hammerblade/torch/kernel/kernel_addmm.cpp
+++ b/hammerblade/torch/kernel/kernel_addmm.cpp
@@ -62,6 +62,7 @@ extern "C" {
                 float sp_result[res_dim_y][res_dim_x];
                 for (int i = 0; i < res_dim_y; i++) {
                     for (int j = 0; j < res_dim_x; j++) {
+                        sp_result[i][j] = 0.0f;
                     }
                 }
 


vb000:
Currently, `malloc` defaults to `Newlib`'s implementation, which allocates memory in 1K chunks. So, it's configured to allocate memory in DRAM (by initializing heap ptr to the end of dram data). If useful, it's totally possible to link with a custom malloc implementation.

yodada (approved):
Looks good! Not sure if @kzheng9 is around during the weekend ...

maybe we can borrow the CS3410 allocator, and use it as our customized scratchpad malloc XD

hammerblade/torch/kernel/kernel_addmm.cpp
```diff
@@ -59,7 +59,7 @@ extern "C" {
                 int res_dim_x = rc == m2_num_blk_per_col - 1 ? m2_last_blk_dim_x : BLOCK_DIM;
 
                 // initialize scratchpad result
-                float sp_result[res_dim_y][res_dim_x] = {};
+                float sp_result[res_dim_y][res_dim_x];
                 for (int i = 0; i < res_dim_y; i++) {
                     for (int j = 0; j < res_dim_x; j++) {
```
sampsyo, line 64: Perhaps missing `sp_result[i][j] = 0.0f`?

sampsyo:
Actually, after looking a little more closely, I do notice that `sp_result` is used uninitialized (via `+=` and `*=` later on). However, there's an existing no-op loop like this:

```
for (int i = 0; i < res_dim_y; i++) {
  for (int j = 0; j < res_dim_x; j++) {
  }
}
```

Maybe that's just missing a proper initializer? There's no rush so we can let @kzheng9 weigh in whenever.

yodada:
hmmmm that uninitialized usage might be the source of #35

sampsyo:
(I added the "manual" initializer line.)

hammerblade/torch/kernel/kernel_addmm.cpp
yodada: Yep! adding this line fixes #35

kzheng9:
Yes, you are right! I thought I had fixed the uninitialized issue because I ran into it for another test case and I was trying to avoid having an explicit loop to initialize `sp_result`, but I did not realize that VLA cannot be initialized with `{}` or `{0}`. Thanks for all the help!

vb000:
Following up on malloc discussion, I'm wondering is it not possible to just use fixed size buffers on scratch pad? Sizes determined by the algorithm...

Since scratchpad is just 1K words, the algorithm has to have strong memory bound checks even with malloc...?

hammerblade/torch/kernel/kernel_addmm.cpp
kzheng9: Actually, I don't think this fixes #35....

kzheng9 (commented):
Actually I take that back, I think this is a necessary fix but this doesn't fix #35.. Looking into it now

sampsyo:
> Following up on malloc discussion, I'm wondering is it not possible to just use fixed size buffers on scratch pad? Sizes determined by the algorithm...

Yeah, that would be pretty cool. The block size should probably be statically determined, and it would be great to check (even at compile time, with [`static_assert`](https://en.cppreference.com/w/cpp/language/static_assert)) that it does not exceed some limit (which would be less than 1K words).

kzheng9:
Please see #50 for a fix to #35. Note that #50 is branched off of this branch so maybe we can combine these two PR's?

sampsyo:
Cool! FWIW, @kzheng9, when one branch is "based off" of another, it is fairly common to just merge them in sequence—we don't necessarily need to combine the PRs first.

PR #45, opened 2020-04-03
Fix 1d reduction corner case

yodada:
BUG fixes:
 - fix the bug in which pytest failures are not communicated to TravisCI correctly
 - fix the bug in which in certain cases, each reduction output is computed with only one input element, and in such condition, the iterator has 1d shape. For example:
```
def test_torch_sum_31():
    x = torch.rand(1, 10)
    _test_torch_sum(x, dim=0)
```


vb000 (approved):
Thanks Lin!

PR #46, opened 2020-04-03
Remove _XOPEN_SOURCE definition to avoid pybind11 compilation error

sampsyo:
In 41bd19e763f86bc405f509f9ee4666e988ee276a, `_XOPEN_SOURCE=500` was introduced to the `CMakeLists.txt` for the PyTorch libraries. This influences the third-party dependency [pybind11](https://github.com/pybind/pybind11/), which uses `strdup`, which is [not a C++ standard function](https://stackoverflow.com/a/12984973/39182)—so that flag enforces stricter standards that remove the function from the standard libraries (on macOS with Clang).

Clang complains, for example:

```
../cmake/../third_party/pybind11/include/pybind11/pybind11.h:193:21: error: use of undeclared identifier 'strdup'
        rec->name = strdup(rec->name ? rec->name : "");
```

I believe these standard-enforcing macros are no longer necessary as of #42, which removes the check for `_XOPEN_SOURCE`. So this PR just backs out that additional flag. (This is why this PR is based on #42—so please merge that first.)

Part of #40.


yodada:
41bd19e763f86bc405f509f9ee4666e988ee276a was added before emulation even exists, so I just gave it try linking against "real" bsg_manycore_cuda library.

```bash
bladerunner/bsg_replicant/libraries/bsg_manycore_features.h:39:2: error: #error "_XOPEN_SOURCE not defined: required for bsg_manycore_runtime"
 #error "_XOPEN_SOURCE not defined: required for bsg_manycore_runtime"
```

I remember Phil telling me `_XOPEN_SOURCE=500` is necessary to get cudalite apps compiled.
@vb000 do you happen to know what that is for?

sampsyo:
Indeed, but that "error" is in some sense self-inflicted. As discussed in #42, the headers include an explicit check for these macros and bail out if they're not enabled—so Phil is right that `-D_XOPEN_SOURCE=500` is "necessary" but only because that check exists. I haven't been able to find a specific place in the code that uses features that actually depend on those source version things, but perhaps @vb000 knows.

yodada:
Ohh right right. They are the same explicit check. I guess we need UW's input on why they are checking for it.

vb000:
Unfortunately, I skipped a level and jumped right from manycore hw to apps ;). I'm learning about runtime as well! I think the whole team is on pytorch slack, I'll start a thread regarding this...

sampsyo:
Hey, y'all—since we haven't quite resolved the underlying question about the manycore libs, but this doesn't seem to be required for emulation (since we got away with #42, after all), would it be OK to merge this? Then I'll be all set up for happy on-laptop development. :smiley:

yodada:
I pushed a commit to the branch, which defines `_XOPEN_SOURCE` if we are *not* using emulation. So we can have emulation compiles fine on both Mac and COSIM. Does that sounds ok? ;)

yodada (approved):

sampsyo:
Perfect! Thanks!!

PR #49, opened 2020-04-04
LeNet-5 on mnist ready to merge

vb000:
After painstaking debug of 3 issues in `addmm`, `max_pool2d` and `sum` respectively, this workload is ready to be merged. In addition to that, following are the changes to current infrastructure:

- Added `[]` operator to HB tensor.
- Use `v1` of `addmm` until solution to https://github.com/cornell-brg/hb-pytorch/issues/35 is ready.
- Fixed a numerical limit bug in `max_pool2d` -- thought `FLT_MIN` means absolute minimum float value, but in fact, it's the lowest positive float value!
- Added a `max_pool2d` testcase to test the consistency of float comparison in `max_pool2d`.

Merge after #50.

### DIFF
--- a/c10/hammerblade/emul/bsg_manycore_cuda.cpp
+++ b/c10/hammerblade/emul/bsg_manycore_cuda.cpp
@@ -40,7 +40,7 @@
 #include <stdio.h>
 #endif
 
-#define DEBUG 1
+#define DEBUG 0
 
 // print out a warning to stderr to remind us that we are
 // running emulation layer
--- a/hammerblade/torch/kernel/bsg_tensor.hpp
+++ b/hammerblade/torch/kernel/bsg_tensor.hpp
@@ -122,11 +122,6 @@ class BSGTensor {
 
       return data[offset];
     }
-
-    DT& operator[](uint32_t n) {
-      bsg_assert_msg(n < N, "error: N=%d but accessed %d\n", N, n);
-      return data[n];
-    }
 };
 
 template<typename T>
--- a/hammerblade/torch/kernel/bsg_tensor.hpp
+++ b/hammerblade/torch/kernel/bsg_tensor.hpp
@@ -122,6 +122,11 @@ class BSGTensor {
 
       return data[offset];
     }
+
+    DT& operator[](uint32_t n) {
+      bsg_assert_msg(n < N, "error: N=%d but accessed %d\n", N, n);
+      return data[n];
+    }
 };
 
 template<typename T>
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,6 @@ install:
     # Install Python requirements
     - pip install --upgrade pip setuptools twine
     - pip install numpy pyyaml mkl mkl-include setuptools cmake cffi typing sklearn tqdm pytest hypothesis
-    - pip install --no-deps torchvision
     - pip list
 
 #------------------------------------------------------------------------
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ install:
     # Install Python requirements
     - pip install --upgrade pip setuptools twine
     - pip install numpy pyyaml mkl mkl-include setuptools cmake cffi typing sklearn tqdm pytest hypothesis
+    - pip install --no-deps torchvision
     - pip list
 
 #------------------------------------------------------------------------
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ This work aims to port PyTorch to HammerBlade.
     `python3.6 -m venv ./venv_pytorch`
  - Install dependencies
     `pip install numpy pyyaml mkl mkl-include setuptools cmake cffi typing sklearn tqdm pytest ninja hypothesis`
+    `pip install --no-deps torchvision`
  - Init pytorch third party dependencies
     `git submodule update --init --recursive`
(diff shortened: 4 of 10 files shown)


c10/hammerblade/emul/bsg_manycore_cuda.cpp
```diff
@@ -40,7 +40,7 @@
 #include <stdio.h>
 #endif
 
-#define DEBUG 1
+#define DEBUG 0
```
yodada, line 43: Yep. It's time to turn it off!

hammerblade/torch/kernel/bsg_tensor.hpp
```diff
@@ -122,6 +122,11 @@ class BSGTensor {
 
       return data[offset];
     }
+
+    DT& operator[](uint32_t n) {
+      bsg_assert_msg(n < N, "error: N=%d but accessed %d\n", N, n);
+      return data[n];
+    }
```
yodada: I guess what you are trying to do here is similar to this case:

https://github.com/cornell-brg/hb-pytorch/blob/8dc12f36bdbc84c4a6b46504b97556f8d1ba7699/hammerblade/torch/kernel/bsg_tensor.hpp#L102-L109

I think @cbatten prefers to use `[]` for tensor accesses. So how about we make the special case , which we assume data is contiguous and index into it with linear indices, `()`, and regular ones `[]`?

yodada (commented):
@kzheng9 so we will switch back to the naive implementation of addmm for now.

yodada:
#44 fixes #35

hammerblade/torch/kernel/bsg_tensor.hpp
vb000: Thought `[]` should be plain C++ array indexing and `()` could be for special indexing? What do you think?

If you still think we should swap definitions, can I do it in a separate pr as it needs modifications to many kernel definitions as well?

vb000:
> #44 fixes #35

Awesome, let's merge it in? I'll try it on this cnn...

vb000:
Oh, sorry, we are waiting for @kzheng9's inputs. Maybe, I could cherry-pick and try the fix...

kzheng9:
Actually I don't think #44 fixes #35..

hammerblade/torch/kernel/bsg_tensor.hpp
sampsyo: FWIW, I would tend to agree with @cbatten's implied suggestion here: use the "right" `operator` override for the job. Parentheses generally imply "calling it like a function" to most programmers, whereas square brackets generally imply "subscripting" to most programmers.

The general aesthetics of operator overloading is that you should be able to use the same "verb" for different "nouns": that is, `foo[bar]` should roughly "mean" the same thing regardless of the type of `foo`. Its implementation is decided by overloading, but the programmer should not have to remember to do different things depending on the type. That's the compiler's job!

vb000: I think it makes sense... 

Even better, I think, we can just have `[]` for indexing. Something like:
```
template<typename... T>
DT& operator[](T... indices) {
  std::initializer_list<uint32_t> iarray = {indices...};

  if (iarray.size() == 1) {
    // implementation for linear indexing: a[5]
  } else {
    // implementation for MDA indexing: a[1, 5]
  }
}
```

kzheng9:
Please see #50 for a fix to #35! Addmm with blocking should be ready to merge (it has been tested with hypothesis :D)

vb000:
Merge after https://github.com/cornell-brg/hb-pytorch/pull/50.

yodada:
> Actually I don't think #44 fixes #35..

The test I used turns out to be too simple ;)

vb000:
I removed `[]` operator from this pr. I'll be creating a separate pr changing `(...)` to `[...]`.

yodada:
> I removed `[]` operator from this pr. I'll be creating a separate pr changing `(...)` to `[...]`.

Got it!

yodada (approved):
LGTM!

hammerblade/torch/kernel/bsg_tensor.hpp
vb000: > I guess what you are trying to do here is similar to this case:
> 
> https://github.com/cornell-brg/hb-pytorch/blob/8dc12f36bdbc84c4a6b46504b97556f8d1ba7699/hammerblade/torch/kernel/bsg_tensor.hpp#L102-L109

Oh, linear indexing is already implemented! Sorry, for some reason, I didn't realize this before

vb000: > I think @cbatten prefers to use `[]` for tensor accesses. So how about we make the special case , which we assume data is contiguous and index into it with linear indices, `()`, and regular ones `[]`?

@cbatten I recall my preference was `[]`, but it's hard to do that in C++.

cbatten: () is not a good operator for indexing. Why is using [] hard to support?

cbatten: I could imagine we could support multidimensional indexing with [] and flattened linear indexing with a method call for example at() ... maybe

cbatten: Adrian is a great C++ hacker. Maybe he has some ideas on how we can use template meta programming to simplify both flattened element wise kernels (like relu) as well as kernels like addmm and sum

PR #51, opened 2020-04-04
Sparse add

zzy82518996:
This pull request is the implementation of performing element-wise add/sub over a dense tensor and a sparse tensor and generate a dense tensor.
Please note that this is not the correct version currently. The reason I create this is because I am currently cannot figure out why the values inside the sparse tensor cannot be correctly read in the kernel. However, the data of values are correctly copied onto the device memory. I will continue to find the problem, but will be thankful that if some of you are available to take a look. Thanks.


zzy82518996:
Now this RP is ready to be merged into the master:
This RP implement the kernel of performing the element-wise add/sub over dense tensor and sparse tensor and generate dense tensor.
We have a new tensorlib_dense_sparse_add HB kernel
Pytorch offload function "add_out_dense_sparse_hb" for tensorlib_dense_sparse_add
data type conversion of indices between CPU and HB.

yodada:
There are many merge conflicts need to be resolved. Also, most of the files changed seems to be moving old tests to another location, which I assume were added by accident ;)

zzy82518996:
Thanks Lin. I think I will close this pull request first and then merge the master to this branch. I will reopen this pull request after I solve the cosim problem of SparseMm kernel.

Issue #54, opened 2020-04-04
Indexing C++ issues


vb000:
I tried using this on [indexing_refactor](https://github.com/cornell-brg/hb-pytorch/tree/indexing_refactor) branch:

https://github.com/cornell-brg/hb-pytorch/blob/d2656d896d0b87105d334f143866792c7df3dcc7/hammerblade/torch/kernel/bsg_tensor.hpp#L87-L88

But seeing this error:
```
               w[co, ci, kh, kw] = 0.0f;
                             ^~
../hammerblade/torch/kernel/kernel_conv.cpp:222:29: warning: left operand of comma operator has no effect [-Wunused-value]
```

I remember this problem before also, my first preference was `[][]..[]` operator, then `[., ., .]`. But I couldn't implement both of these... any suggestions?

sampsyo:
A little Googling suggests that the conventional wisdom for this recommends [currying the subscript operator](https://stackoverflow.com/a/6969904/39182) to allow multiple subscripts like `w[co][ci][kh][kw]`.

Of course, an alternative would be to pass an array of indices, like `w[{co, ci, kh, kw}]`.

vb000:
First option would require recursive calls to the `[]` operator and requires keeping a count of the dimension we are currently in to do that for arbitrary dimensions. After all that, I would have to find a crazy sed command or manually change indexing in different kernels, so implemented the second one :(. Is this acceptable?

vb000:
Also thought about overloading `,` operator to enable `w[co, ci, kh, kw]` (`co`, `ci`, `kh` and `kw` are custom type objects). Not sure if that works, but definitely a potential solution...

cbatten:
The key here is we need to use meta programming. We need to get runtime code for indexing to be super simple. We need the host code to correctly dispatch so that we get a specialized kernel or at least we need to make the common case fast

vb000:
I think the options with indexing we discussed are:

1. Simple indexing code by overloading `()` operator. It's not super nice in terms of readability, but I don't see any technical limitation with this.
2. Indexing by overloading `[]` operator recursively to achieve MDA style (`[][]...[]`) indexing.
3. Indexing by overloading `[]` but requiring user to pass a list instead -- `[{a, b, c}]`.
4. Indexing by overloading `[]` & `,` and defining a custom index datatype to achieve `[a, b, c]` style.

[...]

PR #55, opened 2020-04-05
Indexing refactor

vb000:
Updated HammerBlade Tensor's indexing operator overloading to access tensors using `w[{...}]` instead of `w(...)`.

### DIFF
--- a/hammerblade/torch/kernel/bsg_tensor.hpp
+++ b/hammerblade/torch/kernel/bsg_tensor.hpp
@@ -85,7 +85,7 @@ class BSGTensor {
     }
 
     template<typename... T>
-    DT& operator()(T... indices) {
+    DT& operator[](T... indices) {
       std::initializer_list<uint32_t> iarray = {indices...};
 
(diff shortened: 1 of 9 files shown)


cbatten:
Look. We need to have and actual design discussion about how we want to write clean close in the kernels. Let’s not just have everyone going off and implementing different versions.

yodada:
So what is the plan? Delay this one until we have a formal discussion?

cbatten:
Yes. I think we should discuss and come up with a plan of action for how to: (1) cleanly iterate over HB tensors in device code; and (2) cleanly refactor iteration code into a templated parallel_for style function in such a way that we can support loop unrolling, prefetching, etc

vb000:
> Look. We need to have and actual design discussion about how we want to write clean close in the kernels. Let’s not just have everyone going off and implementing different versions.

We are discussing this here: https://github.com/cornell-brg/hb-pytorch/issues/54. This pr is attempt to address the concern you raised about `()` vs `[]` -- I didn't randomly chose to change it...

PR #56, opened 2020-04-05
Remove serial_foreach in TensorIterator offloading

### LINKED ISSUE #53: Make Iterator offloading more like CUDA
The current way of offloading a `TensorIterator` based operation is very much like CPU. Given what we have learnt in reduction operations, and how things are done on CUDA, I think we would like to make our HammerBlade offloading more CUDA-like

https://github.com/cornell-brg/hb-pytorch/blob/c5b375e43ffeea076f82abbac8972026c5ba3444/aten/src/ATen/native/cuda/Loops.cuh#L135-L190

Basically that means we will give up `iter.serial_foreach`, just like what we did in reductions.

yodada:
In this PR:
 - `brg_*_for` are renamed to `hb_*_for`
 - `hb_*_elementwise_for` now takes in BSGTensors instead of raw pointers
 - `offload_iterator_op_impl` no longer relies on `iter.serial_foreach`
 - now one does not need to write 2 kernels for each reduction


yodada:
Note: trivial case reduction is still not parallelized, but to do so is much simpler than before

aten/src/ATen/native/hammerblade/OffloadIter.cpp
```diff
@@ -19,54 +19,53 @@ void offload_iterator_op_impl(TensorIterator& iter, std::vector<eva_t> device_sc
...
-    for(int i=0; i<device_scalars.size(); i++) {
-      auto alpha = device_scalars[i];
-      device_args.push_back(alpha);
+  // Allocate device tensors and copy the data
+  for(int i=0; i<iter.ntensors(); i++) {
+
+    auto iter_strides = iter.strides(i);
+    // XXX: strides are measured in bytes
+    for(int j=0; j<iter.ndim(); j++) {
+      local_strides[j] = (int64_t)iter_strides[j];
```
vb000, line 46: Think, iterator strides are in bytes to handle other data types like `bfloat16`. Could we just offload word strides so that we can only deal with words on device?

Did the same in reduction offloading... do you think we'd miss anything doing this?
https://github.com/cornell-brg/hb-pytorch/blob/ef8faba6523deb1690a649300991175c7c60e626/aten/src/ATen/native/hammerblade/OffloadIter.h#L54-L57

hammerblade/torch/kernel/hb_elementwise_for.hpp
```diff
@@ -0,0 +1,370 @@
...
+  uint32_t factor = 1;
+  uint32_t offset = 0;
+  for(int32_t i = tensor.ndim() - 1; i >= 0; i--) {
+    uint32_t dimx = idx % sizes[i];
+    offset += dimx * strides[i];
+    idx /= factor;
+    factor *= sizes[i];
+  }
+  return offset;
+}
```
vb000: Can't we just assume tensors in TensorIterators are 1-d tensors (in terms of words)? That saves us lot of index calculation, I believe.

hammerblade/torch/kernel/hb_elementwise_for.hpp
```diff
@@ -0,0 +1,370 @@
...
+
+// =========================================================
+// Tile Element-wise for -- Binary ops
+//
+// This function calculates the per tile range automatically
+//==========================================================
+
+template<typename scalar_t, typename F>
+inline void hb_tile_elementwise_for(BSGTensor<scalar_t> res,
```
vb000: What do think about changing `hb_tile_<>_for` to `hb_parallel_<>_for`? Like the `parallel_for` they defined in pytorch...

vb000 (approved):
LGTM! Just few thoughts going forward...

hammerblade/torch/kernel/hb_elementwise_for.hpp
cbatten: I definitely like hb_parallel_for ... I think we really only need one kind of hb_parallel_for the parallel_for just gives you a linear i and then you can do what you want with that ...

yodada: ok! I also like the name `hb_parallel_<>_for`

hammerblade/torch/kernel/hb_elementwise_for.hpp
yodada: Well, actually no. If a tensor is contiguous in memory, then it can be viewed as 1-d tensor. We handle it with the trivial 1d case here
https://github.com/cornell-brg/hb-pytorch/blob/f52ffeaf665e064fbe42a5513bc6d88717cf72a8/hammerblade/torch/kernel/hb_elementwise_for.hpp#L63-L70
here we can avoid any offset calculation

the problem is tensors that are non-contiguous in memory. Originally iter.serial_foreach takes care of splitting this kind of tensors into subiters (which can be handled by trivial 1d case). Now without serial_foreach, we need to handle them by hand.

aten/src/ATen/native/hammerblade/OffloadIter.cpp
yodada: Yep. I saw your code in reduction offloading. 
I didn't change it in this PR, only because I still haven't figured out why does TensorIterator choose to use bytes as strides measurement. ;)

vb000: I think it's functionally correct because we'd be dividing with `sizeof(scalar_t)`. But yeah, the purpose of byte strides is unclear...

hammerblade/torch/kernel/hb_elementwise_for.hpp
vb000: Cool! I imagine, iterator does this for us: If contiguous, generate 1-d tensors?

PR #57, opened 2020-04-06
mlp_mnist tests

yodada:
Forward and backward tests for mlp mnist model

### DIFF
--- a/hammerblade/torch/tests/test_mlp_mnist.py
+++ b/hammerblade/torch/tests/test_mlp_mnist.py
@@ -7,7 +7,6 @@ import os
 import copy
 import torch
 import torch.nn as nn
-import numpy as np
 import hbutils
 import pytest
 
--- a/hammerblade/torch/tests/test_mlp_mnist.py
+++ b/hammerblade/torch/tests/test_mlp_mnist.py
@@ -20,50 +20,101 @@ torch.manual_seed(42)
 
 class MLPModel(nn.Module):
 
-  def __init__(self):
-    super(MLPModel, self).__init__()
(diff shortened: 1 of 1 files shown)


vb000 (approved):
Looks great!

cbatten:
Does this mean we want to put workloads in the hb-torch repo? Maybe yes? Maybe we should just have a top-level directory called workloads or apps where we put all of our workloads in a clean an stylized way?

yodada:
This one is not a real workload. The tests are only meant for testing the kernels used by mlp mnist. It does not download MNIST dataset, and it does not load any real parameters.

we can certainly put our workloads in the pytorch repo, if that makes more sense.

cbatten:
Ah I see ... I like this idea. We have simple tests in pytorch for all kernels we use in workloads. That way we know the kernels work before we try out the workload ... I think keeping the full workloads (download MNIST, etc) should be in a separate repo ... but I do think every workload should have some corresponding end-to-end tests with simple parameters

yodada:
In the our tests (mlp_mnist and lenet), the models are initialized with random parameters, and we do run them end-to-end. ;)

for example
https://github.com/cornell-brg/hb-pytorch/blob/5aae897bf5d27cd3001e1fc4963238a100a4d310/hammerblade/torch/tests/test_mlp_mnist.py#L88-L92
https://github.com/cornell-brg/hb-pytorch/blob/5aae897bf5d27cd3001e1fc4963238a100a4d310/hammerblade/torch/tests/test_mlp_mnist.py#L102-L104

cbatten:
Perfect!

Issue #61, opened 2020-04-08
logsoftmax failure


yodada:
## 🐛 Bug

Hypothesis found a bug in log_softmax

## To Reproduce

```
Falsifying example: test_log_softmax_hypothesis(
    tensor=array([88.72284], dtype=float32),
)
```

```
>       assert torch.allclose(y, y_hb.cpu(), atol=1e-7)
E       AssertionError: assert False
E        +  where False = <built-in method allclose of type object at 0x7f9ceef7c5e0>(tensor([0.]), tensor([-inf]), atol=1e-07)
E        +    where <built-in method allclose of type object at 0x7f9ceef7c5e0> = torch.allclose
E        +    and   tensor([-inf]) = <built-in method cpu of Tensor object at 0x7f9ce4ffdea0>()
E        +      where <built-in method cpu of Tensor object at 0x7f9ce4ffdea0> = tensor([-inf], device='hammerblade').cpu
```

yodada:
```python
@pytest.mark.skip(reason="known failure")
def test_log_softmax_large_1d():
    x = torch.tensor([88.72284])
    dim = 0
    _test_log_softmax(x, dim)
```

PR #62, opened 2020-04-08
Tests improvements

yodada:
In this PR:
 - import random and set random.seed(42) in *almost* all tests. (https://github.com/pytorch/pytorch/issues/7068)
 - add hypothesis tests to sum/mean/abs/logsoftmax/logsoftmax_backward
 - found a failure in logsoftmax #61


vb000 (approved):
Looks good! Thanks for adding the softmax corner case... on it right away.

PR #63, opened 2020-04-08
Cosim debug

vb000:
Debug flags to cosim. `TRACE=1` generates execution trace and `WAVES=1` generates waveforms.


yodada (approved):
Sweet!

PR #65, opened 2020-04-09
Softmax fix and refactor

### LINKED ISSUE #61: logsoftmax failure

vb000:
- Fix numerical limit issue with large exponent. Improved numerical stability.
- Refactored kernel to use `hb_parallel_for` like other kernels.


yodada (approved):
LGTM! Thanks for the quick fix! ;)

PR #66, opened 2020-04-12
Workaround for HW WAW violation

### LINKED ISSUE #60: kernel_copy_hb_to_hb hangs in COSIM after being called the 3rd time
## 🐛 Bug

Tests hang in COSIM. We traced the bug to `kernel_copy_hb_to_hb`

## To Reproduce
```python
def test_copy_hangs_in_cosim():
    x = torch.ones(10, dtype=torch.long)
    h = x.hammerblade()
    int_h = h.to(torch.int)
    int_h = h.to(torch.int)
    int_h = h.to(torch.int)
```

## Environment

commit 947a1b72b5e892f270ed55f9587daa3ffe4d9b60
bladerunner v4.0.0

## Additional context

We were able to trace it to a missing load in the kernel's inner loop
```asm
 1501   ¦ ¦3d0: 0006a503            lw  x10,0(x13)
 1502   ¦ ¦3d4: 00170713            addi  x14,x14,1
 1503   ¦ ¦3d8: 01e686b3            add x13,x13,x30
 1504   ¦ ¦3dc: 00a5a023            sw  x10,0(x11)
 1505   ¦ ¦3e0: 01f585b3            add x11,x11,x31
 1506   ¦ ¦3e4: fee796e3            bne x15,x14,3d0 <tensorlib_copy_hb_to_hb+0x280>
```
This is the very end of execution trace on COSIM
```
 3391 65321,0,0,3d0,remote_ld_dram
 3392 65322,0,0,3d4,addi
 3393 65323,0,0,3d8,add
 3394 65324,0,0,fffffffc,stall_depend_local_remote_load_dram
 3395 65325,0,0,fffffffc,stall_depend_local_remote_load_dram
 3396 65326,0,0,fffffffc,stall_depend_local_remote_load_dram
 3397 65327,0,0,fffffffc,stall_depend_local_remote_load_dram
 3398 65328,0,0,fffffffc,stall_depend_local_remote_load_dram
 3399 65329,0,0,fffffffc,stall_depend_local_remote_load_dram
 3400 65330,0,0,fffffffc,stall_depend_local_remote_load_dram
 3401 65331,0,0,3dc,remote_st_dram
 3402 65332,0,0,3e0,add
 3403 65333,0,0,3e4,bne
 3404 65335,0,0,3d4,addi
 3405 65336,0,0,3d8,add
 3406 65337,0,0,fffffffc,stall_depend_local_load
 3407 65338,0,0,fffffffc,stall_depend_local_load
 3408 65339,0,0,fffffffc,stall_depend_local_load
 3409 65340,0,0,fffffffc,stall_depend_local_load
 3410 65341,0,0,fffffffc,stall_depend_local_load
 3411 65342,0,0,fffffffc,stall_depend_local_load
 3412 65343,0,0,fffffffc,stall_depend_local_load
 3413 65344,0,0,fffffffc,stall_depend_local_load
 3414 65345,0,0,fffffffc,stall_depend_local_load
 3415 65346,0,0,fffffffc,stall_depend_local_load
 3416 65347,0,0,fffffffc,stall_depend_local_load
 3417 65348,0,0,fffffffc,stall_depend_local_load
 3418 65349,0,0,fffffffc,stall_depend_local_load
 3419 65350,0,0,fffffffc,stall_depend_local_load
 3420 65351,0,0,fffffffc,stall_depend_local_load
```
Note that `3e4,bne` should jump to `3d0,remote_ld_dram`, but it is missing. 
Per @vb000 this could be an issue in the tracing prints. But given we are stalling on dependent load, I think the load is indeed missing.

## Workaround

Do anything that can change the kernel binary, such as adding a `bsg_printf` hides the bug

vb000:
- This adds a method for implementing workarounds for HW issues.
- In this case, this adds a macro to create a RAW hazard to force a dependency. It's used like this:
https://github.com/cornell-brg/hb-pytorch/blob/5d9feb353b7d81b010a199cef9a2c8527e1ac919/hammerblade/torch/kernel/kernel_copy_hb_to_hb.cpp#L21-L25

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -266,4 +266,4 @@ vanilla_operation_trace.csv
 vanilla_stats.csv
 vc_hdrs.h
 vcache_non_blocking_stats.log
-
+DVEfiles
--- a/hammerblade/torch/Makefile
+++ b/hammerblade/torch/Makefile
@@ -50,6 +50,11 @@ regression: kernel.riscv $(COSIM_EXE)
 $(COSIM_EXE):
 	$(MAKE) -C $(COSIM_PYTHON_DIR) $@
 
+ifeq ($(DEBUG),1)
+  RISCV_GXX_EXTRA_OPTS += -g
+  RISCV_LINK_OPTS += -g
+endif
+
 # Include BSG Manycore's builddefs
 include $(BSG_MANYCORE_DIR)/software/mk/Makefile.master
 
--- a/hammerblade/torch/kernel/hb_hw_patch.hpp
+++ b/hammerblade/torch/kernel/hb_hw_patch.hpp
@@ -2,16 +2,16 @@
 #define _HB_HW_PATCH_HPP
 
 #ifndef HB_EMUL
-#define HB_RAW_HAZARD(var) \
-  do {                     \
-    asm volatile (         \
-        "mv %0, %1;"       \
-        : "=r" ((var))     \
-        : "r" ((var))      \
-        );                 \
+#define HB_FIX_WAW_HAZARD(var) \
+  do {                         \
+    asm volatile (             \
+        "mv %0, %1;"           \
+        : "=r" ((var))         \
+        : "r" ((var))          \
+        );                     \
   } while(0)
 #else
-#define HB_RAW_HAZARD(var)
+#define HB_FIX_WAW_HAZARD(var)
 #endif // ifndef HB_EMUL
 
 #endif // _HB_HW_PATCH_HPP
--- a/hammerblade/torch/kernel/hb_hw_patch.hpp
+++ b/hammerblade/torch/kernel/hb_hw_patch.hpp
@@ -1,0 +1,17 @@
+#ifndef _HB_HW_PATCH_HPP
+#define _HB_HW_PATCH_HPP
+
+#ifndef HB_EMUL
+#define HB_RAW_HAZARD(var) \
+  do {                     \
+    asm volatile (         \
+        "mv %0, %1;"       \
+        : "=r" ((var))     \
+        : "r" ((var))      \
+        );                 \
+  } while(0)
+#else
+#define HB_RAW_HAZARD(var)
+#endif // ifndef HB_EMUL
+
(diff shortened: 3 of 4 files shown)


taylor-bsg:
Maybe it would be better to call it more explicitly for what its intent is than what it does? IE HB_FIX_WAW_HAZARD( )

Also in your comments you should explicitly indicate what warning message you should see in cosim when you might need this.

A final comment — is in_size somehow implicitly used by the parallel for, or is this dead code?

vb000:
Actual HW bug is WAW violation. And this tries to fix that by artificially creating a RAW hazard. So `HB_RAW_HAZARD()` adds a RAW hazard? Or we could name it `HB_FIX_WAW_HAZARD()`?

https://github.com/cornell-brg/hb-pytorch/blob/42ae86390522481680dd84f74e303e7e931f5c93/hammerblade/torch/kernel/hb_parallel_for.hpp#L18-L31

`hb_parallel_for` ultimately calls above function which needs sizes. Nevertheless, loading sizes is part of `HBTensor` constructor.

vb000:
Sorry, you suggested fix waw hazard... my bad!

yodada (approved):
Awesome! Thank you so much for working on it!
Just wondering what exactly went wrong in HW?

vb000:
Classic pipeline bug... 'load, load, mul' sequence caused the second load to write during the stall. Unfortunately, first load still in pipeline has overwritten the second.

taylor-bsg:
The bug is a little more obscure than implied by Bandhav’s summary — it
occurs only when you have two non-blocking loads that target the same
output register, and have no intervening use of the output register between
the two instructions (ie the first register value is just thrown away), and
there is a long-stalling instruction (divide or multiply) closely following
the second load. In this case the second register write can be done too
early.

M

vb000:
Yes, it's no way a common bug. Sorry for the over-simplification!

cbatten:
Tricky! Someone should create a new RISCV test that can reproduce this bug so we can have that for all future designs!

PR #64, opened 2020-04-09
Jlw422 sigmoid

jweber97:
sigmoid kernel passes all testing including hypothesis in emul and cosim

PR #68, opened 2020-04-13
Workaround for failures due to comparing floating point results

### LINKED ISSUE #67: Floating point comparison
## ❓ This issue is meant for a discussion on how should we handle floating point comparison.

A few hypothesis / random tests, like `sum_22` and `mean_hypothesis_3d`, may fail. The reason is such:
 - CPU kernels and HB kernels use different algorithms
 - Sometimes the random data includes a mixture of large numbers (like 88 which is favored by hypothesis) and very small numbers (1e-04). 
 - During the `sum` operation, the partial result can become as large as 1000, which creates inaccuracy in the decimals. However, the final result will be small (around 10 or 20). 
 - Then the relative difference between CPU and HB results are large (>1%)

Not sure what is the best way to handle this problem ... We could
 1. Let hypothesis generates small inputs only, this is also the case in upstream
 2. Increase the tolerance

yodada:
Workaround for failures in #64 . The problem is discussed in #67 
In this PR:
 - Hypothesis now generates random values from -2 to 2, instead of -128 to 128.
 - big_rand test in test_addmm uses rand instead of randn.
 - `pytest_runner.py` now sorts the targets list, so the order of execution is fixed.

Warning: this workaround definitely weakens our abilities to spot new bugs.

### DIFF
--- a/hammerblade/torch/tests/hypothesis_test_util.py
+++ b/hammerblade/torch/tests/hypothesis_test_util.py
@@ -25,14 +25,14 @@ def dims(min_value=1, max_value=5):
 def elements_of_type(dtype=np.float32, filter_=None):
     elems = None
     if dtype is np.float32:
-        elems = st.floats(min_value=-128.0, max_value=128.0, width=32)
+        elems = st.floats(min_value=-2.0, max_value=2.0, width=32)
     # elif dtype is np.int32:
     #     elems = st.integers(min_value=0, max_value=2 ** 31 - 1)
     # elif dtype is np.bool:
     #     elems = st.booleans()
     else:
         raise ValueError("Unexpected dtype without elements provided")
-    elems = elems.filter(lambda x: x > 0.0001 or x < -0.0001)
+    elems = elems.filter(lambda x: x > 0.001 or x < -0.001)
     return elems if filter_ is None else elems.filter(filter_)
 
 
--- a/hammerblade/torch/tests/hypothesis_test_util.py
+++ b/hammerblade/torch/tests/hypothesis_test_util.py
@@ -8,7 +8,7 @@ import hypothesis
 import hypothesis.extra.numpy
 import hypothesis.strategies as st
 import numpy as np
-
+import os
 
 def is_travis():
     return 'TRAVIS' in os.environ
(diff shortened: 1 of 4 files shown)


hammerblade/torch/tests/hypothesis_test_util.py
```diff
@@ -25,14 +25,14 @@ def dims(min_value=1, max_value=5):
 def elements_of_type(dtype=np.float32, filter_=None):
     elems = None
     if dtype is np.float32:
-        elems = st.floats(min_value=-128.0, max_value=128.0, width=32)
+        elems = st.floats(min_value=-2.0, max_value=2.0, width=32)
```
vb000: I think larger values are necessary for testing kernels with exponentials (like softmax)... may be we could disable them temporarily.

vb000 (approved):

hammerblade/torch/tests/hypothesis_test_util.py
yodada: That is what I'm worried about as well ... I have no idea what is the best way of handling this

yodada: maybe xfail on travis? http://doc.pytest.org/en/latest/skipping.html#condition-parameter

yodada:
How about this. we use small hypothesis inputs on Travis only.
https://github.com/cornell-brg/hb-pytorch/blob/24d4ad4d98a243e7e3868c8aad2dd90cac34338f/hammerblade/torch/tests/hypothesis_test_util.py#L28-L31

hammerblade/torch/tests/hypothesis_test_util.py
vb000: This sounds good to me.

Actually, just realized adding `rtol` parameter might fix large value mismatches... I'll try that in next the pr.

yodada:
Workaround for failures in #64 . The problem is discussed in #67 
In this PR:
 - when running on Travis, Hypothesis now generates random values from -2 to 2, instead of -128 to 128.
 - big_rand test in test_addmm uses rand instead of randn.
 - `pytest_runner.py` now sorts the targets list, so the order of execution is fixed.

Warning: this workaround definitely weakens our abilities to spot new bugs.

PR #69, opened 2020-04-14
Tensor dim specialization

vb000:
- As a first step of HBTensor method specializations, this pr specializes the low hanging fruit, linear, 0-d and 1-d tensor indexing.
- Some more prefix fixes.
- Increased tolerance of log_softmax from 1e-7 to 1e-6.

p.s: indexing operator `()` will be changed to `.at()` and `.at_safe()` methods in next prs.


yodada (approved):
Great!

PR #72, opened 2020-04-20
"pycosim" as a drop in replacement for "python" with cosim

vb000:
Debugging cosim is hard in itself. Not having full command line interface because of makefile inteface makes the problem worse. So, this pr:

- Installs an executable script, `pycosim`, which attempts to be a drop-in replacement for `python` when running with cosim.
- Installs riscv kernel binary at `<torch_install_prefix>/riscv` during pytorch build to avoid the need for recompiling kernels separately.
- Use arbitrary python cmd line arguments support by VCS cosim binary, `test_loader`, implemented in https://github.com/bespoke-silicon-group/bsg_replicant/pull/558.
- With this, we could run pytest with cosim by:
```
pycosim -m pytest hammerblade/torch/tests/test_add_sub.py
```
- `run-hb-pytest.sh` still works to maintain backward compatibility until the change is released in bsg_bladerunner.

With program size befitting cosim simulation speed, debugging pytest in cosim should now be more or same as that of emulation. `gdb` support is still missing with pytorch cosim though, any ideas how we can incorporate gdb into this approach?


cbatten:
Love it!

yodada (approved):
Awesome!

PR #74, opened 2020-04-21
Adding macro for Unary Ops manually

neiladit:
Some unary operations like `exp` seem to have only CPU invocation from the `UnaryOps.cpp` file. This causes an error of undefined reference during build on adding Hammerblade kernels in `native_functions.yaml` file.
I'm not sure why the corresponding CUDA version of some of these functions is missing. I tried adding a macro to define HammerBlade kernels for `exp` which seems to work.

### DIFF
--- a/aten/src/ATen/native/UnaryOps.cpp
+++ b/aten/src/ATen/native/UnaryOps.cpp
@@ -7,17 +7,17 @@
 #include <ATen/ATen.h>
 #include <ATen/Dispatch.h>
 #include <ATen/ExpandUtils.h>
+#include <ATen/NativeFunctions.h>
 #include <ATen/LegacyTHFunctionsCPU.h>
 #include <ATen/MemoryOverlap.h>
-#include <ATen/NativeFunctions.h>
 #include <ATen/WrapDimUtils.h>
 
 #include <ATen/CPUApplyUtils.h>
-#include <ATen/NamedTensorUtils.h>
 #include <ATen/Parallel.h>
-#include <ATen/core/EnableNamedTensor.h>
-#include <ATen/native/TensorIterator.h>
 #include <ATen/native/UnaryOps.h>
+#include <ATen/native/TensorIterator.h>
+#include <ATen/NamedTensorUtils.h>
+#include <ATen/core/EnableNamedTensor.h>
 
 #include <algorithm>
 #include <cmath>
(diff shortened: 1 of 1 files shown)


sampsyo:
For a little additional context, the way this manifested was that exactly the same mechanism that worked to add a kernel implementation, for example, for `ceil` did not work for `exp`. They seem exactly the same—both unary/in-place operations—what could possibly be different between them?

The answer is that the "glue" for the two functions is defined in different ways. For `ceil` and many others it's this:
https://github.com/cornell-brg/hb-pytorch/blob/8b01234171dc1d16358be6c672256ab176e4d892/aten/src/ATen/native/UnaryOps.cpp#L76-L78

…which works for any target device. For `exp` it's this:
https://github.com/cornell-brg/hb-pytorch/blob/8b01234171dc1d16358be6c672256ab176e4d892/aten/src/ATen/native/UnaryOps.cpp#L316

…which macros seems to hard-code the CPU and CUDA devices. So maybe we're missing something, but if not, that prevents us from adding HammerBlade implementations of any of the kernels in this list?
https://github.com/cornell-brg/hb-pytorch/blob/8b01234171dc1d16358be6c672256ab176e4d892/aten/src/ATen/native/UnaryOps.cpp#L306-L320

Maybe there's a way to do it (or has already been done for one of them), which would be cool. But if not, we may need to modify these macros to enable the HB backend for these kernels.

vb000:
Interesting that `cos` is implemented this way but `sin` is implemented the proper way! Can't think of a reason. 

Will this will be followed by:
```
1116 - func: exp_(Tensor(a!) self) -> Tensor(a!)
1117   supports_named_tensor: True
1118   variants: function, method
1119   dispatch:
1120     CPU: _exp__cpu
1121     CUDA: _exp__cuda
1122     HammerBlade: _exp__hb
```
in `native_functiona.yaml`? That sounds like the way to me.

These two checks in the macro seem to be only difference between two approaches:
https://github.com/cornell-brg/hb-pytorch/blob/e30322b1b449db1ef42016b517dc2a3783acec11/aten/src/ATen/native/UnaryOps.cpp#L288-L295

But why would `cos` need these checks but not `sin`?!

yodada:
I think it's most likely a legacy code issue? I guess it's similar to native dispatching vs all these "stubs".

yodada (approved):

vb000:
Think this is still wip?

neiladit:
@vb000 I did not mean to merge the `exp` function yet. Just wanted to include the change in the `UnaryOps.cpp` file and show an example for `exp`. This change might be required for all these kernels:
https://github.com/cornell-brg/hb-pytorch/blob/8b01234171dc1d16358be6c672256ab176e4d892/aten/src/ATen/native/UnaryOps.cpp#L306-L320

We were planning to do a separate PR for the `exp` kernel if that seems cool.

vb000:
Sounds good! Just wanted to confirm so that we don't accidentally merge, if this is wip...

PR #75, opened 2020-04-22
Cosim is clean

vb000:
We started to fall behind on Cosim with latest hardware. So, I spent past few days improving the cosim execution/debug flow and also fixed a recurring issue. Cosim is running successfully with latest hardware now.

In addition to #72, this pr:
- Adds the WAW fix to `sizes` field in `HBTensor`'s constructor. Load to this particular field is triggering WAW bug in various kernels.
- Install two more executable scripts `pycosim.trace` and `pycosim.wave` to dump 'instruction trace' and 'instruction trace & waveform' respectively. There are intended to be used when we suspect something unexpected is going on in the hardware.

Estimating performance impact of the fix is a nice food for thought =). I expect it to be <10k cycles/batch (or chunk). Reasoning: the fix essentially makes the non-blocking load to `HBTensor.sizes` blocking. And if we assume there are about ~100 kernel invocations instantiating 3-4 tensors on average and each blocking load to dram takes ~20 cycles, lost cycles equals to 100 x 4 x 20 = 8000 cycles. @taylor-bsg what do you think? ;)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -29,7 +29,7 @@ jobs:
           .jenkins/run-shellcheck.sh
       - name: Ensure no tabs
         run: |
-          (! git grep -I -l $'\t' -- . ':(exclude)*.svg' ':(exclude)**Makefile' ':(exclude)**/contrib/**' ':(exclude)third_party' ':(exclude).gitattributes' ':(exclude).gitmodules' ':(exclude)*.mk' || (echo "The above files have tabs; please convert them to spaces"; false))
+          (! git grep -I -l $'\t' -- . ':(exclude)*.svg' ':(exclude)**Makefile' ':(exclude)**/contrib/**' ':(exclude)third_party' ':(exclude).gitattributes' ':(exclude).gitmodules' ':(exclude)*.mk' ':(exclude)*.tcl' || (echo "The above files have tabs; please convert them to spaces"; false))
       - name: Ensure C++ source files are not executable
         run: |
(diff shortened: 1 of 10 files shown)


cbatten:
Can —trace and —wave be command line arguments to pycosim?

vb000:
It's possible, but this way maintains the standard that all cmd line arguments to pycosim go directly to the python interpreter. Moreover, there might be a python module that implements `--trace`. Also, pycosim is a bash wrapper around cosim executable. And bash doesn't have a formal way to support name arguments. Even if we do it with some hacks, there's no good way to document these options in a help message as `--help` would interfere with that of python's `--help`.

cbatten:
Ah ok ... sometimes in these situations people will use a special environment variable to pass configuration options to some kind of "meta thing" ... that is what pypy does. You can configure pypy by setting environment variables and then running pypy

taylor-bsg:
Bandhav, i.e., small overall performance impact? I guess a high-level
percentage estimate would make the scale more clear.

M

vb000:
> Bandhav, i.e., small overall performance impact? I guess a high-level percentage estimate would make the scale more clear. M

Yes, insignificant performance impact. I have't run full workload recently but the inference I demoed took about 200M cycles! Percentage impact would be 10k / 200M = 0.005%. Just an interesting thing to estimate.

The fix is needed in `HBTensor`'s constructor. If its needed in indexing, the impact might get significant. Then we'd think about alternate approaches like confining fixes to only leaf level code, I imagine.

taylor-bsg:
Okay great!

cbatten:
This is a software fix for a hardware bug right? So won’t we be fixing this in the hardware? Or am I getting confused?

vb000:
Apparently, we have to demo PyTorch on the chip we taped out last summer. So, the plan is to not change the hardware (at least temporarily) to avoid divergence...

yodada (approved):

cbatten:
Doing a demo of PyTorch on the chip you taped out last summer sounds awesome! We want to help ... let us know what we can do.

taylor-bsg:
Yes it will be pretty amazing I think!

M

PR #76, opened 2020-04-23
Embedding

yodada:
In this PR:
 - Embedding forward and backward
 - Negation (x = -x)
 - Clamp & Clamp_min & Clamp_max

### DIFF
--- a/hammerblade/torch/kernel/kernel_embedding_backward.cpp
+++ b/hammerblade/torch/kernel/kernel_embedding_backward.cpp
@@ -1,0 +1,54 @@
+//====================================================================
+// Embedding backward kernel
+// 04/22/2020 Lin Cheng (lc873@cornell.edu)
+//====================================================================
+
+#include <kernel_common.hpp>
+
+extern "C" {
+
+  __attribute__ ((noinline))  int tensorlib_embedding_backward(
+          hb_tensor_t* grad_weight_p,
+          hb_tensor_t* grad_p,
+          hb_tensor_t* index_p,
+          int32_t* padding_idx_p,
+          int32_t* num_weights_p,
+          int32_t* numel_p) {
+
+    HBTensor<float> grad_weight(grad_weight_p);
+    HBTensor<float> grad(grad_p);
+    HBTensor<int32_t> index(index_p);
+    int32_t padding_idx = *padding_idx_p;
+    int32_t num_weights = *num_weights_p;
+    int32_t numel = *numel_p;
+    float* grad_weight_data = (float*)grad_weight.data_ptr();
+    float* grad_data = (float*)grad.data_ptr();
+    int32_t* index_data = (int32_t*)index.data_ptr();
+
+    bsg_cuda_print_stat_kernel_start();
+
+    hb_parallel_for(index.numel(), [&](size_t i) {
+      if (index_data[i] != padding_idx) {
+        int32_t k = index_data[i];
+        if (k >= 0 && k < num_weights) {
+          float scale = 1.0;
+          float* dst = grad_weight_data + k * grad_weight.get_strides()[0];
+          float* src = grad_data + i * grad.get_strides()[0];
+          for (size_t j=0; j<numel; j++) {
+            *dst += *src * scale;
+            dst++;
+            src++;
+          }
+        }
+      }
+    });
+
+    bsg_cuda_print_stat_kernel_end();
+    return 0;
+  }
+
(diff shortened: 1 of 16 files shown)


hammerblade/torch/kernel/kernel_index_select.cpp
```diff
@@ -0,0 +1,52 @@
...
+    HBTensor<float> result(result_p);
+    HBTensor<int32_t> index(index_p);
+    float* self_data = (float*)self.data_ptr();
+    float* result_data = (float*)result.data_ptr();
+    int32_t* index_data = (int32_t*)index.data_ptr();
+
+    bsg_cuda_print_stat_kernel_start();
+
+    auto size0 = self.dim(0); // this should really be size(0)
+    auto rowsize = size0 == 0 ? 1 : self.numel() / size0;
```
yodada, line 25: I copied this line from CPU impl, but I don't think it makes much sense ... because if size0 is 0, the tensor has to be tensor([]), and you can't really "select" anything from it. I guess it is here just to prevent divide-by-zero exception.

yodada:
Well ... it seems github's Action service is down again

hammerblade/torch/kernel/kernel_embedding_backward.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+    HBTensor<float> grad_weight(grad_weight_p);
+    HBTensor<float> grad(grad_p);
+    HBTensor<int32_t> index(index_p);
+    int32_t padding_idx = *padding_idx_p;
+    int32_t num_weights = *num_weights_p;
+    int32_t numel = *numel_p;
+    float* grad_weight_data = (float*)grad_weight.data_ptr();
+    float* grad_data = (float*)grad.data_ptr();
+    int32_t* index_data = (int32_t*)index.data_ptr();
```
vb000, line 26: This might be a nitpick... isn't this one of the kernels where we should use raw tensors? Apparently HBTensor-only methods are not called.

hammerblade/torch/kernel/kernel_index_select.cpp
vb000: Wouldn't size be 0 for 0-d tensors (scalars)?

hammerblade/torch/tests/test_embedding.py
```diff
@@ -0,0 +1,38 @@
...
+    else:
+        embedding = nn.Embedding(10, 3)
+    embedding_h = copy.deepcopy(embedding).hammerblade()
+    out = embedding(input)
+    out_h = embedding_h(input.hammerblade())
+    assert out_h.device == torch.device("hammerblade")
+    assert torch.equal(out, out_h.cpu())
+
+def test_torch_nn_embedding_1():
+    input = torch.LongTensor([1, 2, 4, 5])
```
vb000, line 26: Just curious, why do we need LongTensor here?

hammerblade/torch/tests/test_select.py
```diff
@@ -0,0 +1,31 @@
...
+
+import torch
+import random
+
+torch.manual_seed(42)
+random.seed(42)
+
+def _test_select(x, index):
+    h = x.hammerblade()
+    assert h[index].device == torch.device("hammerblade")
```
vb000, line 14: If a select like `x[2,3]` works, throwing in such a test might be useful.

vb000 (approved):

hammerblade/torch/kernel/kernel_index_select.cpp
yodada: there is a check before this one which disallows selecting from 0-d tensors.

hammerblade/torch/tests/test_embedding.py
yodada: Similar to what happens to sparse tensor indices -> there are internal checks on if they are int64 values. Not sure what would break if we change them to int32. For now I downcast them before offloading to HB, but we still takes in int64 in the host code

hammerblade/torch/kernel/kernel_embedding_backward.cpp
yodada: The hope is (in the future) there will be auto-generated wrappers to hide hb_tensor_t interface. Kernel writers will only deal with HBTensors.

hammerblade/torch/tests/test_select.py
yodada: Good point! never thought about that ;)
Added!

PR #77, opened 2020-04-23
addcmul & addcdiv & sqrt

yodada:
Three more kernels in this PR:
 - addcmul
 - addcdiv
 - sqrt


yodada:
addcdiv, addcmul, clamp, embedding, embedding-back, index_select, neg, select, and sqrt are verified on cosim

vb000 (approved):

PR #78, opened 2020-04-27
Kernel Logger

vb000:
- Logs kernel calls as JSON objects during emulation.
- Uses this JSON library for C++: https://github.com/nlohmann/json. Included it as a `hb-pytorch/third_party/include` as header-only library.
- All the logging is done by this class:
https://github.com/cornell-brg/hb-pytorch/blob/06e4377c9f3f46ba4e77c127f453a2398e784c4e/c10/hammerblade/emul/kernel_logger.h#L21-L30
- Log generated for conv2d forward (removed data array as github surprisingly doesn't support vertical scrolling code blocks!):
```json
tensorlib_convolution_forward
[
    {
        "N": 25,
        "data": [
            1.6057322025299072,
            ...
        ],
        "dims": 4,
        "sizes": [
            1,
            1,
            5,
            5
        ],
        "strides": [
            25,
            25,
            5,
            1
        ]
    },
    {
        "N": 25,
        "data": [
            0.13318592309951782,
            ...
        ],
        "dims": 4,
        "sizes": [
            1,
            1,
            5,
            5
        ],
        "strides": [
            25,
            25,
            5,
            1
        ]
    },
    {
        "N": 9,
        "data": [
            0.8822692632675171,
            ...
        ],
        "dims": 4,
        "sizes": [
            1,
            1,
            3,
            3
        ],
        "strides": [
            9,
            9,
            3,
            1
        ]
    },
    {
        "N": 2,
        "data": [
            1.401298464324817e-45,
            1.401298464324817e-45
        ]
    },
    {
        "N": 2,
        "data": [
            1.401298464324817e-45,
            1.401298464324817e-45
        ]
    }
]

### CODE REVIEW

**Comment by yodada** [comment]
I just gave this a try, and it looks like there is a small problem with the current implementation.

This is what I tried
```python
 import torch
 x = torch.tensor([2.])
 x = x.hammerblade()
 y = x.expand(2,2)
 z = y + y
```

This is what being printed
```json
Emulation layer enqueued kernel tensorlib_add

tensorlib_add
[
    {
        "N": 4,
        "data": [
            4.0,
            4.0,
            4.0,
            4.0
        ],
        "dims": 1,
        "sizes": [
            4
        ],
        "strides": [
            4
        ]
    },
    {
        "N": 4,
        "data": [
            2.0,
            0.0,
            1.0700898013745462e-37,
            0.0
        ],
        "dims": 1,
        "sizes": [
            4
        ],
        "strides": [
            0
        ]
    },
    {
        "N": 4,
        "data": [
            2.0,
            0.0,
            1.0700898013745462e-37,
            0.0
        ],
        "dims": 1,
        "sizes": [
            4
        ],
        "strides": [
            0
        ]
    },
    1.0
]
```

This dump does not preserve low level memory layout, and it can potentially trigger segfaults. The problem is, after `expand`, the low level storage does not have length 4 -- it is still the original single element storage instance. If we reconstruct the tensors according to this dump, we would artificially increase the working set, for two reasons: (1) the reconstructed tensor has a larger memory footprint; (2) we reconstruct *two* tensors, instead of one.   

Along with all the metadata (dims, strides, sizes), we need to peak into low level storage info. Here is something I was playing around:
```diff
diff --git a/aten/src/ATen/templates/TensorBody.h b/aten/src/ATen/templates/TensorBody.h
index b9d49b174..9958d90a8 100644
--- a/aten/src/ATen/templates/TensorBody.h
+++ b/aten/src/ATen/templates/TensorBody.h
@@ -278,6 +278,8 @@ class CAFFE2_API Tensor {
   /// Returns a `Tensor`'s device index.
   int64_t get_device() const;

+  int64_t Tensor::get_storage_numel() const;
+
   /// Returns if a `Tensor` has CUDA backend.
   bool is_cuda() const;

diff --git a/aten/src/ATen/templates/TensorMethods.h b/aten/src/ATen/templates/TensorMethods.h
index 0eeea365c..37c289ea9 100644
--- a/aten/src/ATen/templates/TensorMethods.h
+++ b/aten/src/ATen/templates/TensorMethods.h
@@ -79,6 +79,10 @@ inline int64_t Tensor::get_device() const {
   return impl_->get_device();
 }

+inline int64_t Tensor::get_storage_numel() const {
+  return impl_->storage_numel();
+}
+
 inline int64_t get_device(Tensor self) {
   return self.get_device();
 }
diff --git a/c10/core/TensorImpl.h b/c10/core/TensorImpl.h
index 6ed5efa73..026544ee6 100644
--- a/c10/core/TensorImpl.h
+++ b/c10/core/TensorImpl.h
@@ -987,6 +987,10 @@ struct C10_API TensorImpl : public c10::intrusive_ptr_target {

  public:

+  uint64_t storage_numel() {
+      return storage_.numel();
+  }
+
   /**
    * The device type of a Tensor, e.g., DeviceType::CPU or DeviceType::CUDA.
    */
diff --git a/tools/autograd/templates/python_variable_methods.cpp b/tools/autograd/templates/python_variable_methods.cpp
index f9bb57c88..83c772f19 100644
--- a/tools/autograd/templates/python_variable_methods.cpp
+++ b/tools/autograd/templates/python_variable_methods.cpp
@@ -176,6 +176,14 @@ static PyObject * THPVariable_storage_offset(PyObject* self_, PyObject* args)
   END_HANDLE_TH_ERRORS
 }

+static PyObject * THPVariable_storage_numel(PyObject* self_, PyObject* args)
+{
+  HANDLE_TH_ERRORS
+  auto& self = reinterpret_cast<THPVariable*>(self_)->cdata;
+  return wrap(self.get_storage_numel());
+  END_HANDLE_TH_ERRORS
+}
+
 static PyObject * THPVariable_dim(PyObject* self, PyObject* args)
 {
    HANDLE_TH_ERRORS
@@ -915,6 +923,7 @@ PyMethodDef variable_methods[] = {
   {"size", (PyCFunction)(void(*)(void))THPVariable_size, METH_VARARGS | METH_KEYWORDS, NULL},
   {"storage", (PyCFunction)THPVariable_storage, METH_NOARGS, NULL},
   {"storage_offset", (PyCFunction)THPVariable_storage_offset, METH_NOARGS, NULL},
+  {"storage_numel", (PyCFunction)THPVariable_storage_numel, METH_NOARGS, NULL},
   {"storage_type", (PyCFunction)THPVariable_storage_type, METH_NOARGS, NULL},
   {"stride", (PyCFunction)(void(*)(void))THPVariable_stride, METH_VARARGS | METH_KEYWORDS, NULL},
   {"to", (PyCFunction)(void(*)(void))THPVariable_to, METH_VARARGS | METH_KEYWORDS, NULL},
```

The idea is `tensor.imp_->storage_.numel()` gives the actual length of the storage. Combining this info with `elm_size`, `storage_offset`, and `raw_data_ptr`, we can retrieve the lowest level raw memory region. To reconstruct a tensor, we can use `as_strided_tensorimpl`

But there is still this problem of "multiple tensor shares the same storage".

cbatten:
instead of `third_party` maybe we can use `extra`?

yodada:
`third_party` is a directory that exists in upstream pytorch

cbatten:
ah ok ... then that seems like the right place to put this :)

vb000:
Nice catch Lin! Apparently this way of data access only handles contiguous tensors correctly. For fixing this, I think we don't have to look at metadata of native tensors, because all the metadata we need to access the device tensor should be in the device tensor? 

I'll try to fix this using strides to replicate the memory layout. `N` doesn't correspond to memory layout but `strides` does! Am I still missing something?

vb000:
I implemented a fix to handle all possible dense tensors... the logic is that we can always calculate the size of a dimension `d` in the memory from the `sizes[d]` unless `strides[d] == 0`. In which case, size of dimension d in memory is 1. Can't think of any scenario breaking this... your thoughts?

Your approach is simpler and more elegant, but needs us to add `storage_numel` field to device tensor's metadata. I'm okay doing that as well.

yodada:
Ahha! It's not too hard to think of a counter example ;) But it's not related to your calculation. Your claim holds. 

The problem comes from this line
https://github.com/cornell-brg/hb-pytorch/blob/06e4377c9f3f46ba4e77c127f453a2398e784c4e/c10/hammerblade/emul/kernel_logger.h#L126

In certain cases, you will be able to find out the *actual* number of elements. However, it may not be linear in memory. Consider this example:
```python
x = torch.tensor([[1,2,3],[4,5,6],[7,8,9]])
```
Here the memory layout is
```bash
|1 2 3 4 5 6 7 8 9|
```
Then we do the following:
```python
x = x.t()[0]
```
`x` becomes `tensor([1,4,7])`
and we need to read from
```bash
| 1 2 3 4 5 6 7 8 9 |
  ^     ^     ^
```
but the code above does a linear read which gives you 1, 2, 3

vb000:
In the example, `storage_numel` of second tensor `x` = 9?

vb000:
I think I got the issue now... multiple tensors can refer to a storage element. And there's no other way to reproduce the exact storage layout than to look at `storage_numel`?

yodada:
Correct. In this example, `x` which has shape (3,3), `x'` which is `x.t()`, and `x''` which is `x.t()[0]` they all have the same *storage* under the hood.

```bash
x    x'   x''
|    |    |
\    |    /
  storage
 (0x0-0x8)
```

yodada:
And yet we still have this "multiple reference to single storage" problem.

yodada:
If we naively reconstruct tensors, without considering they could refer to the same storage, we are in trouble.

vb000:
An idea... let's append pointers to tensor arguments in the json object? We can chop off MSBs of data pointers in the emulation and let cosim consider resulting addresses as EVAs. This has an added benefit of reproducing the distance between arguments in the dram, if device memory allocated in emulation is a single contiguous block.

vb000:
Nevertheless, I think we could tackle inaccuracy in modeling step by step. This step makes the improvement of running 'key kernel calls with exact data' to the higher-level weak scaling approach. There are corner cases to be handled such as non-contiguous in-place kernel calls, which I think are not part of the common case (key kernels).

Anyway this PR is the first step of the problem: implementing a method to log kernel calls. There is no infrastructure to run the json and verify if the modeling is accurate yet. That'll be part of the second step, so let's defer the discussion of modeling accuracy to that?

yodada:
Oh the emulation malloc is no longer a naive linear allocator actually. It’s just a wrapper around vanilla malloc.

Here is what I’ve been trying (not much luck yet) — we dump a “computational graph” which tells you which kernel is called with what input size in the ROI. Then we can have a way to specify which one of these dump kernels you want to run in cosim. The computation will be performed on CPU until we reach that kernel. Then we analyze the inputs (at most 4 or 5 tensors I guess?), perform low level storage copy, reconstruct metadata, and invoke this very kernel on cosim. This way we can still have the choice of running kernels we are interested in in parallel.

Low level storage copy and metadata reconstruction is simple, and I know what to do, but redispatching to HB is very tricky ...

vb000:
Yes, re-dispatching is tricky and I think, we don't really need to do that. The idea is to use the storage  copied to the json file and run key kernels in it, in cosim.

vb000:
My understanding of tensor storage concisely is, each tensor refers to a chuck of data in the memory which possibly could be referred by other tensors as well. Within that chunk, each tensor has a `storage_offset` pointing to the first element of this tensor. 

To reconstruct a tensor exactly, all we need is the 'storage element's head pointer', 'storage offset' and 'contents of the storage'. Added these fields to the json log so that we could reconstruct the tensors exactly for cosim.

yodada (approved):
Looks good! One small concern is the code works for float type tensors only. Certain loss function takes in integer tensors, but one may argue loss functions take only a few percent of total execution time and we may never study them in isolation

vb000:
That's a limitation. The problem was, when logging, there's no good way to figure out what the data type of the tensor kernel argument is...

PR #79, opened 2020-04-28
eye kernel with tests

mchao409:
Added eye kernel (identity)

### DIFF
--- a/aten/src/ATen/native/hammerblade/Eye.cpp
+++ b/aten/src/ATen/native/hammerblade/Eye.cpp
@@ -13,11 +13,7 @@ namespace native {
       m = n;
     }
     output.resize_({n,m});
-    device_args.push_back(create_device_tensor(output, device_ptrs));
-    device_args.push_back(create_device_scalar(n));
-    device_args.push_back(create_device_scalar(m));
-    c10::hammerblade::offload_kernel("tensorlib_eye", device_args);
-    cleanup_device(device_args, device_ptrs);
+    hb_offload_kernel(output, n, m, "tensorlib_eye"); 
     return output;
   }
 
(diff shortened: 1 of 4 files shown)


hammerblade/torch/kernel/kernel_eye.cpp
```diff
@@ -0,0 +1,31 @@
+#include <kernel_common.hpp>
+
+
+extern "C" {
+
+  __attribute__ ((noinline)) int tensorlib_eye(
+    hb_tensor_t* output, long* n, long* m) {
+    auto y = HBTensor<float>(output);
+    auto N = *n;
```
cbatten: I general we should not abuse auto. The type `long` is the same number of characters as `auto`. So this should probably be:
```
HBTensor<float> y(output);
long N = *n;
```
The above is more explicit, more compact, and more readable. Just a thought :)
```

File: aten/src/ATen/native/hammerblade/Eye.cpp [hunk 2]
```diff
@@ -0,0 +1,25 @@
...
+namespace native {
+
+  Tensor& eye_out_hb(Tensor& output, long n, long m) {
+    std::vector<eva_t> device_args;
+    std::vector<eva_t> device_ptrs;
+    if(m == -1) {
+      m = n;
+    }
+    output.resize_({n,m});
+    device_args.push_back(create_device_tensor(output, device_ptrs));
```
yodada: I think you don't have to call these low level APIs yourself? `hb_offload_kernel` will take care of them for you.
You can write
```C
hb_offload_kernel(output, n, m, "tensorlib_eye");
```

hammerblade/torch/kernel/kernel_eye.cpp
mchao409: Just made the change

aten/src/ATen/native/hammerblade/Eye.cpp
mchao409: Fixed!

yodada (approved):
LGTM! Thanks!

PR #81, opened 2020-05-01
Propagate storage as void pointer for logging

vb000:
This makes sure that log won't break when the tensor has data other than float.


yodada (approved):
Cool!

PR #82, opened 2020-05-01
Low-level Copy

yodada:
This PR implements `llcopy` and its Python binding `hammerblade_ll`. This operator copies the low level storage space to HB and reconstructs a HB tensor from it. See `hammerblade/torch/tests/test_llcopy.py` for more details.

Note: this operator *does not* prevent double copying issue for now. It will be implemented in the next PR


vb000 (approved):
Looks good! Is it possible to add to add a docstring to python function so that user can know what this does by `__doc__`? Also, I think it's worth mentioning that this may not create a contiguous tensor.

yodada:
Good points! I'll add docs about it

yodada:
LLCopy has been verified on COSIM

PR #83, opened 2020-05-01
Python interface for kernel call logger

vb000:
- Added a flag to check if hb is emulated, from python.
- Logs kernel calls to a file
- Enable/disable interface to python


yodada:
According to the error message, it seems you are trying including a ATen file in c10. That's won't work since torch depends on c10 and they are built separately (c10 before ATen).

vb000:
Thanks Lin! I think we can include some files, but in this case I tried to include a generated header. As you said, since c10 and ATen are built concurrently, the generated file was missing. Because it was generated in my previous build, the build was fine in my local copy!

Anyway, those were just for some debugging I forgot to remove ;).

hammerblade/torch/tests/test_mlp_mnist.py
```diff
@@ -39,7 +38,7 @@ def forward(self, x):
 # Forward pass
 # -------------------------------------------------------------------------
 
-@pytest.mark.skipif(os.environ.get('USE_HB_EMUL') is None, reason="Slow on cosim")
+@pytest.mark.skipif(torch.hb_emul_on, reason="Slow on cosim")
```
yodada, line 41: Really like this!

yodada (approved):
LGTM!

hammerblade/torch/tests/test_mlp_mnist.py
vb000: Yes, should avoid using build time env variables at run time.

PR #86, opened 2020-05-04
support pip install

yodada:
This PR contains a few tricks to allow
```
pip install git+https://github.com/cornell-brg/hb-pytorch.git
```

 - added a method `setup_emul` to `setup.py`, which sets a few ENV
 - if `setup.py` is called with `egg_info`, do not build torch. We use this pass to install dependencies like pyyaml.
 - if `setup.py` is called with `--record`, we trigger torch building.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -386,7 +386,8 @@ def build_deps():
 ################################################################################
 
 # the list of runtime dependencies required by this built package
-install_requires = []
+install_requires = ['numpy', 'requests', 'six', 'sklearn', 'tqdm',
+                    'pytest', 'ninja', 'hypothesis']
 
 if sys.version_info <= (2, 7):
     install_requires += ['future', 'typing']
--- a/setup.py
+++ b/setup.py
@@ -202,6 +202,9 @@ EMIT_BUILD_WARNING = False
 RERUN_CMAKE = False
 CMAKE_ONLY = False
 filtered_args = []
+# setup emulation environment variables if requested
+SETUP_EMUL = False
+
 for i, arg in enumerate(sys.argv):
     if arg == '--cmake':
         RERUN_CMAKE = True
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -282,6 +292,30 @@ def report(*args):
...
+    os.environ["USE_QNNPACK"] = "0"
+    os.environ["USE_DISTRIBUTED"] = "0"
+    os.environ["USE_OPENMP"] = "0"
+    os.environ["ATEN_THREADING"] = "NATIVE"
+    os.environ["OMP_NUM_THREADS"] = "1"
+    os.environ["USE_HB_EMUL"] = "1"
+    if os.getenv("BSG_MANYCORE_DIR", "") != "":
+        del os.environ["BSG_MANYCORE_DIR"]
+    if os.getenv("HB_KERNEL_DIR", "") != "":
+        del os.environ["HB_KERNEL_DIR"]
```
vb000, line 317: Couldn't understand this part... where do we set these variables?

yodada: Precautions. Same as
https://github.com/cornell-brg/hb-pytorch/blob/413850923789b97b317edf43585d62e6a50d119e/setup_emul_build_env.sh#L29-L31

vb000: Oh, sorry, this is emulation, we don't need them... my bad!

setup.py
```diff
@@ -352,7 +386,9 @@ def check_file(f):
 ################################################################################
 
 # the list of runtime dependencies required by this built package
-install_requires = []
+install_requires = ['numpy', 'requests', 'six', 'sklearn', 'tqdm',
+                    'pytest', 'ninja', 'hypothesis', 'pyyaml',
+                    'typing']
```
vb000, line 391: I wonder why pytorch didn't do this!

vb000 (approved):
Great!

PR #71, opened 2020-04-15
Native profiling tools

yodada:
In this PR:
 - native execution time breakdown
 - while running on x86, unimplemented kernel profiling tool performs shadow HB kernel lookup to discover necessary but unimplemented kernels for running on HB
 - explicitly set USE_MKL=0
 - updated README on how to use native profiling tools

https://docs.google.com/presentation/d/1A595LHyjirgo8DWNv6dPpMxLmI0L5ghKilc5W2LKXjg/edit#slide=id.p

PR #87, opened 2020-05-04
HB Profiler v0

vb000:
Implements HB profiler and python bindings to profile individual kernels from python.

A tiny program like this:
```python
  1 import torch
  2 import torch.hammerblade.profiler as hbprof
  3
  4 x = torch.rand(2, 3).hammerblade()
  5 y = torch.rand(2, 3).hammerblade()
  6
  7 hbprof.enable()
  8 print(x + y)
  9 print(torch.mm(x, y.t()))
 10 print(hbprof.summary())
```

Outputs this on cosim:
```
tensor([[1.0914, 0.4428, 0.4938],
        [1.0323, 1.3454, 0.2439]], device='hammerblade')
tensor([[0.2552, 0.4545],
        [0.2490, 0.5537]], device='hammerblade')

-----------------+-------------+------------------+------------------+
Kernel Name      | HB Total %  | HB Total Cycles  | Number of Calls  |
-----------------+-------------+------------------+------------------+
tensorlib_add    | 24.4588     | 7M               | 1                |
tensorlib_addmm  | 50.3176     | 15M              | 1                |
tensorlib_fill   | 25.2236     | 7M               | 1                |
-----------------+-------------+------------------+------------------+

BSG REGRESSION TEST PASSED
BSG COSIM PASS: Test passed!
```
Also works on emulation, but everything except number of calls is 0. For complete interface, see https://github.com/cornell-brg/hb-pytorch/tree/cosim_profiler#execution-summary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,37 @@ torch.aten_profiler.unimpl_print()
 
 ### HB Profiling
 
+#### Execution summary
+HammerBlade offloading api also supports profiling the execution time using CUDALite api functions. This can be used as:
+
+```python
+import torch
+import torch.hammerblade.profiler as hbprof
+
+x = torch.rand(2, 3).hammerblade()
+y = torch.rand(2, 3).hammerblade()
+
+# Enables profiling
+hbprof.enable()
+print(x + y)
+print(x * y)
+
+# Prints summary of add and mul kernels
+print(hbprof.summary())
+
+# Disbales profiling: div kernel below won't be profiled
+hbprof.disable()
+print(x / y)
+
+# Clears profiling data structure
+hbprof.clear()
+
+# Enables profiling and only profiles mm kernel
+hbprof.enable()
+print(torch.mm(x, y.t()))
+print(hbprof.summary()
+```
+
 #### HB Kernel Call Logs
 HB emulation can output a file with the list of kernel calls along with associated data in json format. This can be used as:
 
--- a/c10/hammerblade/CMakeLists.txt
+++ b/c10/hammerblade/CMakeLists.txt
@@ -83,6 +83,11 @@ else()
     "-DHB_KERNEL_PATH=${CMAKE_INSTALL_PREFIX}/riscv/kernel.riscv"
     )
 
+  target_compile_definitions(
+    c10_hammerblade PRIVATE
+    "-DCOSIM"
+    )
+
   target_include_directories( c10_hammerblade PUBLIC $ENV{BSG_MANYCORE_DIR} )
 endif()
 
--- a/c10/hammerblade/HammerBladeFunctions.cpp
+++ b/c10/hammerblade/HammerBladeFunctions.cpp
@@ -3,10 +3,10 @@
 
 #include <mutex>
 
+c10::hammerblade::HBProfiler hb_profiler __attribute__ ((visibility ("default")));
+
 namespace c10 {
 namespace hammerblade {
-
-HBProfiler hb_profiler;
 
 namespace {
 static std::once_flag init_flag;
--- a/c10/hammerblade/HammerBladeFunctions.cpp
+++ b/c10/hammerblade/HammerBladeFunctions.cpp
@@ -1,9 +1,12 @@
 #include "HammerBladeFunctions.h"
+#include "HammerBladeProfiler.h"
 
 #include <mutex>
 
 namespace c10 {
 namespace hammerblade {
+
+HBProfiler hb_profiler;
 
 namespace {
 static std::once_flag init_flag;
@@ -72,7 +75,10 @@ void offload_kernel(const char* kernel, std::vector<eva_t> args) {
 
   C10_HB_CHECK(hb_mc_kernel_enqueue(&_hb_device, _hb_grid_dim, _hb_tg_dim, kernel,
                                     args.size(), cuda_argv));
+
+  hb_profiler.kernel_start(kernel);
   C10_HB_CHECK(hb_mc_device_tile_groups_execute(&_hb_device));
+  hb_profiler.kernel_end(kernel);
 
   free(cuda_argv);
 }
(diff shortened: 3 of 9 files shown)


cbatten:
This is awesome! How does this work? How do you get the cycle count from python?

cbatten:
@yodada already has an API to turn on/off profiling natively. We should probably use the same API for both so it just works regardless of whether we are running it natively or on hbcosim?

vb000:
> This is awesome! How does this work? How do you get the cycle count from python?

This uses `bsg_time()` CUDALite api function to get the simulation time before and after the offloading DPI call. `bsg_time() / 10` is the cycle count. This estimate is probably pessimistic. Dustin and Borna are coming up with more precise function that also accounts for overheads.

cbatten:
OK! Yes, I think `bsg_time()` is probably not the exact API we want, but this is a great proof of concept.

vb000:
Yes, and this could be easily updated to use the new api by just changing couple of methods in the HBProfiler class.

vb000:
> @yodada already has an API to turn on/off profiling natively. We should probably use the same API for both so it just works regardless of whether we are running it natively or on hbcosim?

All the interaction with CUDALite happens through `c10`, whereas native profiling has to happen in `aten`. Basically we have to snoop into different libraries, and consequently two separate classes.

cbatten:
I am talking about the python API. `hbprof.enable()` we should have a single way from the python workload to turn on profiling and it should "just work" for both native and HB profiling.

vb000:
Oh I see! That sounds great! And, have two `summary`s? `hbprof.native_summary()` and `hbprof.hb_summary()`?

cbatten:
Just one summary that outputs both native and HB stats?

cbatten:
Some kernels might be native and some kernels might be on hbcosim.

vb000:
Cool! One table with both CPU and HB kernels with different set of columns for CPU to output time as opposed to number of cycles?

Moreover we can estimate HB time assuming 1 Ghz clock.

cbatten:
I think we should output cycles, seconds, and dynamic instructions for both native and HBcosim if possible

vb000:
Dynamic instruction count on HB might need a new api function. That'd be nice addition!

vb000:
In fact, I'll check if we can expose C++ api to all profiling tools we now have...

vb000:
I checked how I can integrate this with aten_profiler (#89?), it seems to be have it's own c10 snooper and is tightly coupled with kernel re-dispatching between CPU and HB. Also there are differences in formatting, formatting sounds minor but is time consuming. So, it might be hard integrate both. @yodada Do you agree?

The high-level approach of this pr is same as that of the pytorch's builtin `torch.autograd.profiler`, so I'm thinking I could extend that to use this. What do you think? @cbatten

cbatten:
All I am saying is that we have a single API from python to turn on and off both profilers. Both profilers can have completely different implementations ...

vb000:
I understand, aggregating output into a single table is hard. May be as a first step, whenever user requests to output profiled data, the interface can print two tables, one for native kernels and other for HB kernels?

vb000:
One solution could be both C++ profilers return json objects to python and a common python profiler can aggregate the data when returning to user. @yodada Does that sound good?

cbatten:
Two tables is perfectly fine :) Doesn't need to be one table ... I was mostly just talking about the Python API from the workload perspective.

yodada:
oh #89 actually meant to make `ATenProfiler` a general thing. The idea is you can "plugin" different components to it. Now it has `Execution time`, `Unimp`, `(Execution) Chart` and `Route (redispatching)`. You can remove any one of them and others won't be affected at all.

yodada:
`ATenProfilerLog` provides 2 events (1) entering a func, and (2) returns from a func. A component can choose to hijack one of them, or both. For example, `ExecutionTime` logs start time at entering, and store the delta at returning. `Chart` only logs which func you touch at entering.

yodada:
`ATenProfiler` provides 4 API calls
 - enable()
 - disable()
 - is_in_roi()
 - is_top_level()

Any other functionality of profiling tools are implemented by "components"

yodada:
Another example is here
https://github.com/cornell-brg/hb-pytorch/blob/5c0e93208f6db6df3548eb9b73a341d0deeaf816/c10/hammerblade/HammerBladeFunctions.cpp#L64-L67
By creating a ATenProfilerLog in `offload_kernel`, we roughly captures the execution on device

vb000:
I thought following features in this pr are super useful, to generalize aten_profiler as the standard profiler, could you also include these?

- Profiler should be compiled by default to avoid requiring multiple builds -- profiler should always be available to be enabled at runtime.
- Ability to clear the profiler and restart it multiple times in a single program.
- Logs a table with execution stats per device kernel.
- Profiler should output a string so that user can decide whether to print or write to file.
- Also, we could just call it `profiler`?

You might want to reuse routines in this file: https://github.com/cornell-brg/hb-pytorch/blob/103b78b6bca0c19e6761c0ac6c3d894279aaf9ce/c10/hammerblade/HammerBladeProfiler.cpp

yodada:
> I thought following features in this pr are super useful, to generalize aten_profiler as the standard profiler, could you also include these?
> 
> * Profiler should be compiled by default to avoid requiring multiple builds -- profiler should always be available to be enabled at runtime.
this equals to set `PROFILE_ATEN` to `ON` by default in CMakeLists.txt
> * Ability to clear the profiler and restart it multiple times in a single program.
we have this since #71 
> * Logs a table with execution stats per device kernel.
device kernels are included but not as first class citizens. should be simple to add
> * Profiler should output a string so that user can decide whether to print or write to file.
we have this in #89 
> * Also, we could just call it `profiler`?
ok!
> 
> You might want to reuse routines in this file: https://github.com/cornell-brg/hb-pytorch/blob/103b78b6bca0c19e6761c0ac6c3d894279aaf9ce/c10/hammerblade/HammerBladeProfiler.cpp

vb000:
Closing this as Lin has implemented something similar with aten_profiler.

PR #88, opened 2020-05-05
kz73_addmm_unroll

kzheng9:
Performance improvements for the addmm kernel. Includes:
- `BLOCK_SIZE` is now a factor of 2 (8)
- refactor `dram_to_sp` and `compute` to make code more readable
    - also allows experimenting with different unroll factors
- separate common case vs general case when computing a block to minimize number of branch instructions
    - common case: result block size is same as `BLOCK_SIZE` (8)
    - general case: result block size can be smaller than `BLOCK_SIZE` -> partial blocks require additional checks when loading from DRAM and computing
- unroll by 8 in innerloop of both `dram_to_sp` and `compute`
    - this seems like the best unroll factor based on preliminary testing (may need more investigation though!)

Future TODOs:
- investigate best unroll factor
- more microbenchmarking on loading data from DRAM to scratchpad

### DIFF
--- a/hammerblade/torch/kernel/kernel_addmm.cpp
+++ b/hammerblade/torch/kernel/kernel_addmm.cpp
@@ -5,9 +5,60 @@
 
 #include <kernel_common.hpp>
 #include <cmath>
-#define BLOCK_DIM 10 // sqrt(4KB/4 byte/4 data matrix) = 15 max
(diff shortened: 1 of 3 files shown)


yodada (approved):
I think it looks great!

kzheng9:
Ok! I don't think I have access to merge, could you please help me merge it in? Thanks!

yodada:
Done!

PR #89, opened 2020-05-05
HBProfiler Cleanup

yodada:
In this PR:
 - Profiling related code is moved to `c10/probe`, which is built by default.
 - Top level profiling is renamed to `HBProfiler`
 - Profiling functionalities are moved into different components
 -  `ExecutionTime` -> how long does each kernel take
 - `Unimplemented` -> which kernel hasn't been implemented on HB yet
 - `Chart` -> add "kernel-of-interest" by `add_beacon`, then `Chart` will generates an "execution chart" -- a sequence of kernel-of-interest.
 - `Route` -> if `redispatching` is enabled (by set option `HB_REDISPACH`), `route` takes in a json which has the same as what has been produced by `Chart`.
 - Printing related APIs in https://github.com/cornell-brg/hb-pytorch/blob/lc873-05-04-ATP-cleanup/torch/hb_profiler.py return string by default


yodada:
- `ATenProfiler` rename to `HBProfiler`
 - APIs defined in `HBprofiler.h` have been cleaned up
 - Execution time fancy prints now returns a string, instead of printing the results directly

TODO:
 - Cleanup naming of component APIs

yodada:
Ok now the only leftover is redispatching logic, which is turned on by default

yodada:
I think this PR is ... finally ready to be reviewed ... 
I have to admit that it's a little bit too long ...

vb000 (approved):
Great set of features!  Only pain point I felt is, it's little complicated to get a conceptual understanding of what all is possible. Is it possible to show recommended usage using a toy program? Thanks!

yodada:
How to use the profiling tool is mostly captured in https://github.com/cornell-brg/hb-pytorch/blob/lc873-05-04-ATP-cleanup/hammerblade/torch/tests/test_profiling.py 
But my interoperation of your request is about how to add a new "component" to this *ugly* (I arguably went too far in terms of make it looks cleaner) framework?  


Let me try to explain how did I add `Chart`, which logs which "key kernels" are executed in a run. The idea is user to need register one or more `beacons`, which are the signatures of kernels we want to log. It is named beacons since ... you know ... chart and beacons are both navigation terms (also `route` and `waypoints` for redispatching)

The first thing we do is add `c10/probe/Chart.h` (https://github.com/cornell-brg/hb-pytorch/blob/lc873-05-04-ATP-cleanup/c10/probe/Chart.h). The idea is each component will have a class, and a global instance. 
https://github.com/cornell-brg/hb-pytorch/blob/0903b3bf9a699312f99c37dec3f40c97aad1de74/c10/probe/Chart.h#L12-L24
https://github.com/cornell-brg/hb-pytorch/blob/0903b3bf9a699312f99c37dec3f40c97aad1de74/c10/probe/Chart.h#L31

We also declares `C10_PROBE_APIs`
https://github.com/cornell-brg/hb-pytorch/blob/0903b3bf9a699312f99c37dec3f40c97aad1de74/c10/probe/Chart.h#L26-L29
These API func simply calls into the global instance. For example, `C10_PROBE_API const std::string chart_print()` simply calls `g_execution_charter.print()`
https://github.com/cornell-brg/hb-pytorch/blob/0903b3bf9a699312f99c37dec3f40c97aad1de74/c10/probe/Chart.cpp#L63-L65

`ExecutionCharter` has only 1 functionality during profiling -- look at each kernel, if the kernel is at "top-level" (call stack depth == 1) and it is a "key kernel", we log it to a json. To do so, we add a call to its `log` function in the "kernel entering event" provided by `HBProfiler`
https://github.com/cornell-brg/hb-pytorch/blob/0903b3bf9a699312f99c37dec3f40c97aad1de74/c10/probe/HBProfiler.cpp#L83-L92
Specifically, we make the call if the current kernel is "top-level"
https://github.com/cornell-brg/hb-pytorch/blob/0903b3bf9a699312f99c37dec3f40c97aad1de74/c10/probe/HBProfiler.cpp#L88-L89
and checks if it is a *key kernel*  inside `log`
https://github.com/cornell-brg/hb-pytorch/blob/0903b3bf9a699312f99c37dec3f40c97aad1de74/c10/probe/Chart.cpp#L19-L23

Hope this can help!

yodada:
A usage example of `Chart` can be find here
https://github.com/cornell-brg/hb-pytorch/blob/0903b3bf9a699312f99c37dec3f40c97aad1de74/hammerblade/torch/tests/test_profiling.py#L70-L81

yodada:
Oh @vb000 I have a thought regarding python API -- how about we make `hb_profiler.py` a top level thing, and it only does two things (1) enable/disable; (2) imports other methods from components.

For example, currently we do `torch.hb_profiler.exec_time_raw_stack()`. I think it should really be
`torch.hb_profiler.exec_time.raw_stack()`. So things are more modular and organized I guess?

yodada:
Moved 
https://github.com/cornell-brg/hb-pytorch/blob/51e38c954fb30ef688a0b0286ae29f08fce6b73f/torch/hb_profiler/__init__.py#L1-L26
will be included in the next PR

vb000:
I'm not sure about begin required to import a module for each stat. If the profiler has all the data available, why not provide a function like `torch.hb_profiler.stats()` to return formatted stats and user can just print them?

A minor thing, I believe python supports namespaces through module and submodules. We could just put all hammerblade submodules under `torch.hammerblade` namespace, than having to prefix them `hb_`...?

yodada:
I think I simply want to prevent hb_profiler from getting too long and too complex. I can image we have a top level method `stats()` which will do the data gathering from each components.

That is good point. Moving everything to hammerblade ...

vb000:
> A usage example of `Chart` can be find here
> https://github.com/cornell-brg/hb-pytorch/blob/0903b3bf9a699312f99c37dec3f40c97aad1de74/hammerblade/torch/tests/test_profiling.py#L70-L81

Great, thanks! Jsons are great for communicating between programs but I'm not sure it'd be super convenient for us to manually write them, does re-dispatching really need json objects and not just a list of function names?

yodada:
Oh I feel you, definitely! My original thought was the user needs `Chart` to tell you the sequence of "key kernels", and it dumps a json. Something like this
```json
[
    {
        "offload": false,
        "signature": "at::Tensor at::CPUType::{anonymous}::addmm(const at::Tensor&, const at::Tensor&, const at::Tensor&, c10::Scalar, c10::Scalar)"
    },
    {
        "offload": false,
        "signature": "at::Tensor at::CPUType::{anonymous}::addmm(const at::Tensor&, const at::Tensor&, const at::Tensor&, c10::Scalar, c10::Scalar)"
    }
]
```
This is exactly what `Route` takes in. One just need to change that `offload` field to true.

yodada:
So it's more like `Chart` gives you the json, and you only need to turn the switches ;)

yodada:
A few changes mentioned above are included in #91

PR #92, opened 2020-05-09
DMA tensor copy

yodada:
In this PR:
 - Emulate DMA APIs
 - Move tensor data around with DMA -- but we still do memcpy on metadata and scalars.


vb000 (approved):
Looks great!

PR #93, opened 2020-05-09
Pass device ptrs by reference

vb000:
- Passes device ptrs by reference
- Prints a single emulation message at device init and remove enque messages.


yodada (approved):
Cool!

PR #94, opened 2020-05-09
[emul] make sure there is no concurrent tile group execution

yodada:
Although it is very unlikely ... but added a check to make sure we won't try executing more than one tile group with HB emul

### DIFF
--- a/c10/hammerblade/emul/kernel_trampoline.cpp
+++ b/c10/hammerblade/emul/kernel_trampoline.cpp
@@ -1,5 +1,8 @@
-#include <iostream>
+#include <c10/util/Exception.h>
 #include <kernel_trampoline.h>
+
+#include <iostream>
+#include <atomic>
 
 std::map<std::string, std::function<int(uint32_t, uint64_t*)>> kernelMap;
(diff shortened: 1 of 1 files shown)


c10/hammerblade/emul/kernel_trampoline.cpp
```diff
@@ -36,10 +43,20 @@ int execute_kernels() {
     return HB_MC_FAIL;
   }
 
+  int idle = IDLE;
+  if (!hb_device_status.compare_exchange_strong(idle, IN_USE)) {
+    TORCH_CHECK(false, "HB device is already in use")
+  }
+
```
vb000, line 49:
```cpp
TORCH_CHECK(
  hb_device_status.compare_exchange_strong(idle, IN_USE), 
  "HB device is already in use")
```
;)

vb000 (approved):

c10/hammerblade/emul/kernel_trampoline.cpp
yodada: done!

PR #95, opened 2020-05-09
Move concurrency check to offload kernel

yodada:
Re-think about #94 ... it would be better to do the check in `offload_kernel`


vb000 (approved):

vb000:
Yeah, this need not be emulation specific!

yodada:
Originally I thought the error I posted on slack today is somewhat related to "having multiple threads trying executing HB kernels" ... so I added this check ... but it now seems unrelated ...

PR #97, opened 2020-05-13
[emul] bring back emulation layer debug print

yodada:
This PR brings back the `Emulation layer enqueued kernel xxxx` prints, but only if set `EMUL_DEBUG` environment variable.


yodada:
wait what??? the flake8 failures are not even in the changeset ... we never touched them ...

vb000 (approved):
Great! This is more convenient I guess! We could enable/disable messages with:

```
$ <EMUL_DEBUG=1> python setup.py install
```

vb000:
Might be a good idea to mention this easter egg in the readme...

cbatten:
Maybe all environment variables that control either emulation or cosim should have some kind of prefix to avoid namespace collisions? Maybe `HBEMUL_` or `HBCOSIM_` ? so it would be `HBEMUL_DEBUG=1` or like `HBCOSIM_LOG_FILE=dirforlogfiles` stuff like that?

yodada:
Ok! We can do `HBEMUL_DEBUG=1`

PR #98, opened 2020-05-14
fix building without HB

yodada:
Fixed a bug in which llcopy prevents pytorch from building correctly when USE_HB is set to OFF.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,9 @@ This work aims to port PyTorch to HammerBlade.
 
       python setup.py develop
 
+- Turn on emulation debug info
+      export HBEMUL_DEBUG=1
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -51,6 +51,9 @@ This work aims to port PyTorch to HammerBlade.
 
       python setup.py develop
 
+- Turn on emulation debug info
+      export HBEMUL_DEBUG=1
+
```
yodada, line 56: not the correct PR ... but we do we have it doc'ed ;)

vb000: Cool!

vb000 (approved):

PR #99, opened 2020-05-14
fix flake8 errors

yodada:
Somehow there are new flake8 errors popped up


yodada:
I have no idea why these cases didn't pop up before ...

vb000 (approved):

PR #100, opened 2020-05-14
Fix logical linear indexing to nD conversion

yodada:
This appears to be a very fundamental bug ... in which logical linear to actual offset conversion was done incorrectly. not sure how did we get away from it ...

The conversion logic is similar to
https://stackoverflow.com/questions/12429492/how-to-convert-a-monodimensional-index-to-corresponding-indices-in-a-multidimens

### DIFF
--- a/hammerblade/torch/kernel/hb_parallel_for.hpp
+++ b/hammerblade/torch/kernel/hb_parallel_for.hpp
@@ -22,9 +22,9 @@ inline uint32_t offset_calc(uint32_t idx, HBTensor<scalar_t> tensor) {
   uint32_t factor = 1;
   uint32_t offset = 0;
   for(int32_t i = tensor.ndim() - 1; i >= 0; i--) {
+    idx /= factor;
     uint32_t dimx = idx % sizes[i];
     offset += dimx * strides[i];
-    idx /= factor;
     factor *= sizes[i];
   }
   return offset;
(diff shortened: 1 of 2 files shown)


hammerblade/torch/kernel/hb_parallel_for.hpp
```diff
@@ -22,9 +22,8 @@ inline uint32_t offset_calc(uint32_t idx, HBTensor<scalar_t> tensor) {
   uint32_t factor = 1;
   uint32_t offset = 0;
   for(int32_t i = tensor.ndim() - 1; i >= 0; i--) {
-    uint32_t dimx = idx % sizes[i];
+    uint32_t dimx = (idx / factor) % sizes[i];
     offset += dimx * strides[i];
-    idx /= factor;
     factor *= sizes[i];
```
vb000: Cool! In many programs though, I've seen semantics like below for index calculation (`%` paired with `/`), so might be easily recognized and understood:
```
    uint32_t dimx = idx % sizes[i];
    idx /= sizes[i];
    offset += dimx * strides[i];
```

vb000 (approved):

hammerblade/torch/kernel/hb_parallel_for.hpp
vb000: Actually, we save multiplies with above code? But probably, compiler is smart enough to remove the extra multiplies in the original code...

yodada: ok! one fewer integer multiplication! updated the code.
just make sure I'm doing it correctly ;)

vb000: Really?! Compiler didn't optimize it out?

vb000: Looks good to me btw!

yodada: oh I didn't check assembly ;)
yeah I assume compilers are smart

vb000: Agree, if we can fix it in source, why bother...

PR #101, opened 2020-05-15
Back to the future (CPU)

yodada:
Continue execution on CPU after re-dispatching to HB, so we can do result comparison. 
One may also offload a number of kernels in a single run.

### DIFF
--- a/aten/src/ATen/function_wrapper.py
+++ b/aten/src/ATen/function_wrapper.py
@@ -1389,7 +1389,7 @@ def create_generic(top_env, declarations):
                         ret_elements.append(returns[r]['name'])
                     else:
                         ret_elements.append("std::get<{0}>(res)".format(r))
-                ret_call = option['return_type'] + "(" + ",".join(ret_elements) + ");"
+                ret_call = 'return ' + option['return_type'] + "(" + ",".join(ret_elements) + ");"
                 option['redispatch_ret_call'] = ret_call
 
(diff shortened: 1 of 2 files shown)


vb000:
https://github.com/cornell-brg/hb-pytorch/blob/38032598c2359995339f2105a943cc7a3047ddb4/hammerblade/torch/tests/test_profiling.py#L128-L134

If profiler is enabled we redispatch, otherwise we run on cpu... did I understand it correctly?

aten/src/ATen/function_wrapper.py
```diff
@@ -1331,6 +1334,7 @@ def process_redispatch_cpu_to_hb(option):
             # type: (FunctionOption) -> None
             redispatch_condition = "(c10::probe::use_hb_redispatch() && c10::probe::hb_profiler_is_in_roi() && c10::probe::hb_profiler_is_top_level() && c10::probe::should_redispatch(__PRETTY_FUNCTION__))"
```
yodada, line 1335: Here is the `redispatching` condition:
(1) built with redispatching
(2) profiler is enabled
(3) is a top level kernel
(4) `Route` json says this kernel should be redispatched
There is also a hidden condition -- the cpu is currently running on CPU

vb000: Oh okay, thanks for pointing it out!

vb000 (approved):

PR #103, opened 2020-05-18
Fallback to CPU automatically if an ATen OP is not implemented for HB

yodada:
If a certain ATen OP is not implemented for HB, one would have the choice to fallback to CPU.
Normal execution on HB can be resumed after falling back to CPU.

BUG fix:
 - llcopy now has a backward method -- which does nothing

### DIFF
--- a/aten/src/ATen/function_wrapper.py
+++ b/aten/src/ATen/function_wrapper.py
@@ -1452,7 +1452,7 @@ TensorList {0}_hb = TensorList({0}_hb_vec);
         def process_fallback_hb_to_cpu(option):
             # type: (FunctionOption) -> None
             option['fallback_warning'] = "std::cerr << \"ATen OP {0} falls back to CPU\" << std::endl;\n".format(option['api_name'])
-            fallback_condition = "(c10::probe::use_hb_redispatch() && c10::probe::hb_profiler_is_in_roi() && c10::probe::hb_profiler_is_top_level())"
+            fallback_condition = "(c10::probe::use_hb_redispatch() && c10::probe::hb_profiler_is_in_roi() && c10::probe::hb_profiler_is_top_level() && c10::probe::fallback_is_enabled())"
             fallback_boxing = ""
             fallback_unboxing = ""
             fallback_actuals = []
@@ -2104,7 +2104,7 @@ TensorList {0}_hb = TensorList({0}_hb_vec);
     def process_fallback_hb_to_cpu(option):
         # type: (FunctionOption) -> None
         option['fallback_warning'] = "std::cerr << \"ATen OP {0} falls back to CPU\" << std::endl;\n".format(option['api_name'])
-        fallback_condition = "(c10::probe::use_hb_redispatch() && c10::probe::hb_profiler_is_in_roi() && c10::probe::hb_profiler_is_top_level())"
+        fallback_condition = "(c10::probe::use_hb_redispatch() && c10::probe::hb_profiler_is_in_roi() && c10::probe::hb_profiler_is_top_level() && c10::probe::fallback_is_enabled())"
         fallback_boxing = ""
         fallback_unboxing = ""
         fallback_actuals = []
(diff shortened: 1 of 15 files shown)


yodada:
Let me give a summary of the CPU fallback logic
 - for typedefault operators -- we do a try-catch to capture HB related exceptions. If the exception is due to a kernel / stub not been implemented on HB, we re-run the operator on CPU
- for backend operators -- we simply re-dispatch to CPU op

hammerblade/torch/tests/test_fallback.py
```diff
@@ -0,0 +1,71 @@
...
+
+def test_complex_1():
+    input = torch.randn(3)
+    target = torch.empty(3).random_(2)
+    cpu_loss = F.binary_cross_entropy_with_logits(input, target)
+    torch.hammerblade.profiler.enable()
+    torch.hammerblade.profiler.fallback.enable()
+    hb_loss = F.binary_cross_entropy_with_logits(input.hammerblade(), target.hammerblade())
+    torch.hammerblade.profiler.fallback.disable()
+    torch.hammerblade.profiler.disable()
```
yodada, line 62: This is an example of how to use `Fallback` to run `binary_cross_entropy_with_logits`

aten/src/ATen/function_wrapper.py
```diff
@@ -1991,6 +2103,75 @@ def process_redispatch_cpu_to_hb(option):
         assert alter_method_dispatch
         option['alter_method_dispatch'] = alter_method_dispatch
 
+    def process_fallback_hb_to_cpu(option):
+        # type: (FunctionOption) -> None
+        option['fallback_warning'] = "std::cerr << \"ATen OP {0} falls back to CPU\" << std::endl;\n".format(option['api_name'])
+        fallback_condition = "(c10::probe::use_hb_redispatch() && c10::probe::hb_profiler_is_in_roi() && c10::probe::hb_profiler_is_top_level() && c10::probe::fallback_is_enabled())"
```
yodada, line 2109: Here is the fallback condition
 - redispatching is enabled
 - in profiler ROI
 - is top level op
 - fallback is enabled
There are 2 hidden conditons:
 - currently running on HB
 - for general ops, the exception must be HB related.

aten/src/ATen/function_wrapper.py
```diff
@@ -122,7 +122,21 @@ def TypedDict(name, attrs, total=True):  # type: ignore
     ${device_guard_declaration}
 #ifdef HB_REDISPATCH
     if(!${redispatch_condition}) {
-      ${return_call} at::native::${native_type_method_dispatch}(${native_actuals});
+      try {
+        ${return_call} at::native::${native_type_method_dispatch}(${native_actuals});
+      } catch(const c10::Error& c10_error) {
+        if(${fallback_condition} &&
+          (c10_error.msg_without_backtrace().find("missing HammerBlade kernel") != std::string::npos
+           || c10_error.msg_without_backtrace().find("is not implemented for HB, and Fallback is not enabled") != std::string::npos)) {
```
yodada, line 130: Here is one of the hidden conditions -- the fallback-able exception must be HB related -- missing stub or HB backend

yodada:
After this PR, I'll take a detour to look at multiple tiles. 
I'm pretty happy with our workload profiling / evaluation toolset, though there are 2 missing pieces:
 - double copying
 - API to read cosim stats

yodada:
Oh right -- there is one case where, even with this PR, we still cannot run a certain workload end-to-end: if we have a partially implemented HB operator.

For example -- our `add` works with `float32` only. `BatchNorm` requires adding 2 `long`.

vb000 (approved):
Great! Any reason why we'd want this be tied with profiler?

yodada:
Semantically I think it shouldn't be tied to the profiler, but in practice it relies on a piece of info from the profiler -- if an operator is at top level. So here it is

PR #105, opened 2020-05-21
HBTensor partial template specialization

vb000:
- Separated HBTensor's implementation to HBTensorImpl class, so that HBTensor can be partially specialized without code duplication.
- HBTensor is now a derived class of HBTensorImpl
- HBTensor is partially specialized with `dims` parameter, and using this, strides are cached on scratchpad.
- Following are the speed ups observed on an 8x8 micro-benchmark:
   - Conv2d: 1.24x
   - Addmm: 1.8x
- Fixed a bug in pytest skipif

### DIFF
--- a/hammerblade/torch/kernel/hb_tensor.hpp
+++ b/hammerblade/torch/kernel/hb_tensor.hpp
@@ -61,8 +61,106 @@ typedef struct {
 // allocation.
 // =========================================================
 
-template <typename DT>
+template <typename DT, uint32_t dims=-1>
 class HBTensor {
+  private:
+    uint32_t N;
+    uint32_t strides[dims];
+    uint32_t sizes[dims];
+    DT* data;
+
+  public:
+    HBTensor(hb_tensor_t* t) :
+      N(t->N),
+      data((DT*) ((intptr_t) t->data)) {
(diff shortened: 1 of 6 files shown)


vb000:
In some discussions at UW, there were some ideas to use scratchpad as a software cache. I realize, that idea is particularly relevant to HBTensor class as tensor accesses can be cached across last two dimensions. 

If there exists a mechanism to cache tensor accesses that is generally applicable to all kernels, memory optimizations can be abstracted out into HBTensor class and kernels can solely concentrate on algorithmic optimizations. So, I'm also exploring a mechanism to cache data accesses in HBTensor class. Curious to see how it goes!

cbatten:
I think the problem is if there is no reuse then it does not make sense to copy the data into the scratchpad ... however, if there is frequently access tensor metadata then copying that into the scratchpad might make sense?

vb000:
This PR handles just the metadata part: strides. Strides are universally accessed by every kernel instantiating tensors. And strides in most cases only take up 4 words -- lot of bang for the buck!

Regarding data accesses though, I'm not sure as well. I'll start by analyzing access patterns of kernels. But I think, HBTensor class can abstract out not only caching, but also prefetching of data!

yodada (approved):
LGTM!

PR #106, opened 2020-05-26
multiple-tile kernels + multiple-tile emulation

yodada:
In this PR:
 - A std::thread based multi-tile emulation layer. User can control the *emulated device* size by setting environment variable `HBEMUL_TILE_X_DIM` and `HBEMUL_TILE_Y_DIM`. By default, the machine size is 1x1. Note: the assumption is we will use a tile group size == machine size.
 - When using a 1x1 machine, the behavior is identical to original single tile impl.
 - (naively) parallelized all kernels we have, or made them thread-safe. Not much can be done on these element-wise kernels, but kernels such as addmm and conv2d need more work.
 - Makes `eye` better.

Known issues:
 - Dot is broken -- working on a fix


vb000 (approved):
Looks great! One minor mismatch I guess is with global variables in kernel_<>.cpp files... by default, they have to be considered as thread local, unless explicitly attributed to `.dram` section.

yodada:
Yes! I realized that after looking at the linker script -- they are actually thread local variables. That is something we really need to keep in mind

yodada:
One more thing, this PR assumes a tiny modification to UW's cpp barrier
```diff
diff --git a/software/bsg_manycore_lib/bsg_tile_group_barrier.hpp b/software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
index 8952588..dda7e1a 100644
--- a/software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
+++ b/software/bsg_manycore_lib/bsg_tile_group_barrier.hpp
@@ -230,6 +230,11 @@ public:
         this->c_barrier.reset();
     }

+    void reset() {
+        this->r_barrier.reset();
+        this->c_barrier.reset();
+    }
+
     //  The main sync funciton
     void sync() {
         // If barrier dimensions is 1x1, i.e. only a single tile is
```

PR #110, opened 2020-05-29
Updated changed cosim target

vb000:
The target to build sim executables changed in bsg_replicant. This pr updates the setup cosim script to just run a small testcase to build sim executables -- the interface for which is apparently more stable.


yodada (approved):
LGTM!

PR #113, opened 2020-06-08
Include $(BSG_F1_DIR)/machine.mk

yodada:
Include $(BSG_F1_DIR)/machine.mk in riscv.mk so we can get the correct machine dimensions.


yodada:
Hmmm. another error in untouched files ...

vb000 (approved):

PR #114, opened 2020-06-08
Perf Eval Flow

yodada:
In this PR:
 - per ATen OP data dump
 - scripts for post-processing

How things work -- https://docs.google.com/presentation/d/18ViM9BNCS4YdUZyOIW9IY8EwFyrWJ85X7bWGcQEtvko/edit?usp=sharing


c10/hammerblade/HammerBladeFunctions.cpp
```diff
@@ -14,6 +14,10 @@ const int IDLE = -1;
 const int IN_USE = -42;
 std::atomic<int> hb_device_status{IDLE};
 
+// bsg_time to cycles conversion factor
+// check #87
+const uint64_t cycle_time = 10000;
```
vb000, line 19: Isn't this 400000?

vb000 (approved):
To be honest, I'm so unfamiliar with this region of PyTorch that, I'd need more time to understand the details. So, I'm mostly going with the presentation and the doc for now, and possibly provide late feedback when I actually work with this. Running a chunk and assuming weak scaling sounds great (at least for the deadline)!

Basically, we'd calculate time it takes to execute this on hammerblade
```
python mypytorchapp.py --nepoch=1 --nbatch=1 --bathc_size=<chunk size>
```

and compare it with the time it takes to execute this on CPU. Correct?
```
python mypytorchapp.py --nepoch=1 --nbatch=1 --bathc_size=<full batch size>
```

yodada:
See #122

PR #115, opened 2020-06-09
Clang support

vb000:
Past few days I've been working on supporting non-blocking loads with LLVM. Initial results are very promising. This adds support for kernel compilation with Clang using related updates in bsg_manycore builddefs.

When installing hb-pytorch for cosim, kernels can be chosen to be compiled with Clang by:
```
$ CLANG=1 python setup.py develop
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,18 +7,41 @@ This work aims to port PyTorch to HammerBlade.
 
 ### How to build PyTorch to use COSIM
   This assumes that you have a working COSIM installed. Then you can either put `hb-pytorch` under `bsg_bladerunner`, or set `BRG_BSG_BLADERUNNER_DIR` to your `bladerunner` path.
- - Clone hb-pytorch repo
-    `git clone -b hb-device git@github.com:cornell-brg/hb-pytorch.git`
- - Create python virtual environment
-    `python3.6 -m venv ./venv_pytorch`
- - Install dependencies
-    `pip install numpy pyyaml mkl mkl-include setuptools cmake cffi typing sklearn tqdm pytest ninja hypothesis`
- - Init pytorch third party dependencies
-    `git submodule update --init --recursive`
- - Setup building environment variables.
-    `cd hb-pytorch && source setup_cosim_build_env.sh`
- - Build pytorch. This step can take up to 15 minutes
-    `python setup.py develop`
+ - Enable devtoolset-8 or any toolchain that supports C++14.
+ - Set following variable to point to `bsg_bladerunner` clone:
+ ```
+ export BRG_BSG_BLADERUNNER_DIR=<path to bsg_bladerunner that has be setup>
+ ```
+ - Clone hb-pytorch repo:
+ ```
+ git clone -b hb-device git@github.com:cornell-brg/hb-pytorch.git
+ ```
+ - Create python virtual environment:
+ ```
+ python3.6 -m venv ./venv_pytorch`
+ ```
+ - Install dependencies:
+ ```
+ pip install numpy pyyaml mkl mkl-include setuptools cmake cffi typing sklearn tqdm pytest ninja hypothesis
+ ```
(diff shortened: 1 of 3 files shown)


sampsyo:
Sorry if this is a dumb question, but I don't quite see where the compiler is actually switched to Clang to compile the kernels… maybe the README also needs to recommend something like `CXX=/path/to/clang++` or something?

yodada:
If I understand the logic correctly, the magic happens here
https://github.com/cornell-brg/hb-pytorch/blob/5e4a70a4f7fa5e84896ef95ded0490ed95ca5dbb/hammerblade/torch/riscv.mk#L59
then when the bsg Makefiles are included
https://github.com/cornell-brg/hb-pytorch/blob/5e4a70a4f7fa5e84896ef95ded0490ed95ca5dbb/hammerblade/torch/riscv.mk#L64

which leads us to 

https://github.com/bespoke-silicon-group/bsg_manycore/blob/7226108af28b34e60d05bca985c3c02a9d72cc60/software/mk/Makefile.builddefs#L189-L199

sampsyo:
Got it; thanks! Looks like `Makefile.builddefs` has some complex stuff that replaces GCC-based rules with some Clang-based rules to do multi-step compilation with several separate commands (C to LLVM IR, run a custom pass on the LLVM IR, compile LLVM IR to assembly text with `llc`, use GCC to compile assembly to an object file). Interesting!

yodada (approved):
LGTM!

vb000:
Sorry, should have put in the description so that you didn't have to reverse engineer! Also, we use GCC compiled libraries: https://github.com/bespoke-silicon-group/bsg_manycore/blob/7226108af28b34e60d05bca985c3c02a9d72cc60/software/mk/Makefile.builddefs#L220-L222/.

The commands use intrinsic cross-compilation capability of LLVM, and as a result work with system llvm (> 10.0.0)! But, system LLVM won't have our changes to cost-model and scheduler. So, to get better performance we need to use our fork of llvm.

taylor-bsg:
Over the long run, we would definitely like a single-executable drop in
replacement for gcc so that we can use the compiler (without our special
makefiles) in OPM (other people’s makefiles).

vb000:
We kind of have to do multi-step compilation to be able to add out-of-source custom passes. Unless, we stipulate that custom passes have to be in-llvm-source and should be added to our llvm fork.

Providing makefrags to help cross-compilation for manycore seems like a natural usecase of `make`, and think it's useful. In fact, I think we should even provide similar CMake cross-compilation file in `bsg_manycore`, so that we don't have to do ugly 'make in cmake' hacks, like the one I did here.

taylor-bsg:
Yes, I think we would just have our own llvm fork.

Or we can wrap the passes with an python or bash driver.

For example, consider recompiling PanicRoom or other modules that are
packaged with their own makefile infrastructure and just expect the user to
change CC and some parameter flags. We had this discussion extensively last
year — forcing everybody to fit inside our own Makefile methodology is very
limiting for portability of new infrastructures.


M

vb000:
Sounds great! Yeah, I agree with elegant drivers, and in fact, we have a nice example in `pycosim` that tries to provide seamless experience of running python on cosim using a bash driver.

sampsyo:
Here is how to run an out-of-tree LLVM pass as part of a single Clang invocation:
https://www.cs.cornell.edu/~asampson/blog/clangpass.html

vb000:
> Here is how to run an out-of-tree LLVM pass as part of a single Clang invocation:
> https://www.cs.cornell.edu/~asampson/blog/clangpass.html

For well supported targets this might be perfect! But for relatively new targets like RISCV, choosing a subtarget (`-mcpu`/`-mtune`) for example, doesn't seem to work from Clang driver. Some small issues like these might need us to use multi-step compilation...?

PR #116, opened 2020-06-09
Type casting between int, long, float, and double

### LINKED ISSUE #24: On Device Type Casting
## 🐛 Bug

Type casting on HammerBlade works for long -> int only.

## To Reproduce

This test demos the bug
```
def test_torch_to_int_to_long_1():
  x = torch.ones(10, dtype=torch.int)
  assert x.type() == 'torch.IntTensor'
  h = x.hammerblade()
  long_x = x.to(torch.long)
  ong_h = h.to(torch.long)
  assert long_h.type() == 'torch.hammerblade.LongTensor'
  assert long_h.device == torch.device("hammerblade")
  assert torch.equal(long_x, long_h.cpu())
```

yodada:
In this PR:
 - added the functionality that allows us to convert tensors between int, long, float, and double.
    this partially resolves #24 (finally)


vb000 (approved):
Great! This is super convenient!

PR #119, opened 2020-06-16
log and exp_ kernels

yodada:
In this PR:
 - registered exp_ op
 - added log_out kernel
 - with these 2 kernels, we can run `BinaryCrossEntropyWithLogits` on HB

### DIFF
--- a/hammerblade/torch/tests/test_log.py
+++ b/hammerblade/torch/tests/test_log.py
@@ -1,0 +1,36 @@
+"""
+Tests for log
+06/16/2020 Lin Cheng (lc873@cornell.edu)
+"""
+import torch
+import random
+from hypothesis import given, settings
+from .hypothesis_test_util import HypothesisUtil as hu
+
+torch.manual_seed(42)
+random.seed(42)
+
+def _test_torch_log(x):
+    h = x.hammerblade()
+    log_x = x.log()
+    log_h = h.log()
+    assert log_h.device == torch.device("hammerblade")
+    assert torch.allclose(log_h.cpu(), log_x, equal_nan=True)
+
+def test_torch_log_1():
+    x = torch.randn(3, 5)
+    _test_torch_log(x)
+
+def test_torch_log_2():
+    x = torch.ones(10)
+    _test_torch_log(x)
+
+def test_torch_log_3():
+    x = torch.zeros(5)
+    _test_torch_log(x)
+
+@settings(deadline=None)
+@given(tensor=hu.tensor())
+def test_torch_log_hypothesis(tensor):
+    x = torch.tensor(tensor)
+    _test_torch_log(x)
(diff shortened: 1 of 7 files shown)


hammerblade/torch/tests/test_log.py
```diff
@@ -0,0 +1,36 @@
...
+    x = torch.ones(10)
+    _test_torch_log(x)
+
+def test_torch_log_3():
+    x = torch.zeros(5)
+    _test_torch_log(x)
+
+@settings(deadline=None)
+@given(tensor=hu.tensor())
+def test_torch_log_hypothesis(tensor):
```
vb000, line 34: Hypothesis tests take extremely long on cosim... each test runs about >50 testcases, I guess. At least for cosim, should we manually figure out some corner cases and write them as individual tests? Not a problem for emulation of course.

vb000 (approved):
Looks good!

hammerblade/torch/tests/test_log.py
yodada: In my experience hypothesis runs for roughly 100 tests if no issue is found. The problem is really when one of them failed. Hypothesis will try to "shrink" it -- find another failure case where the input is shorter/simpler. this process can really take time on cosim.

vb000: Yeah, I think we need to make a compromise here. Cosim regression taking so long might actually become a deterrent to making sure cosim regression is clean. Since the emulation is parallel, it's a pretty good indicator of correctness, I think. So, let's run hypothesis tests only in emulation?

yodada: To be honest I always run cosim tests with `-k-hypothesis` flag, which excludes all hypothesis tests  ;)

cbatten: If we want to disable hypothesis for cosim then we will need more directed tests to make sure things are working on cosim ... mostly because we need to make sure there are no bugs in the cosim RTL

vb000: Yes, that requires little more effort in finding corner cases. Especially, we would have to look floating point stuff, small/large numbers, zero/infinity, NaNs etc. And also with corner case dimensions: single element tensor etc. I think adding these conner cases would be turn out to be quicker than running 100 tests on cosim!

cbatten: actually I have a better idea ... why don't we just run like 5 cases on cosim? hypothesis will try a different 5 cases each time ... I think we should combine that with some better directed testing too ... having at least a little random testing on cosim seems like a good idea? so we can do 100 examples on emulation and 5 examples on cosim ... there might be way we can automate that so hypothesis/pytest automatically choses the right number of examples

taylor-bsg:
Neat idea. Aside: Can any of this be done on FPGA or the silicon to speed
it up?

M

cbatten:
Do you mean for testing? Definitely! I think Bandhav had examples of pytest/pytorch working with F1 right? Using F1 to do more regressions of the hardware would be very cool.

If you mean for workload performance analysis then yes -- I think if we can get some kind of reasonable performance model on F1 then we could also take advantage of that using our pytest/pytorch infrastructure. If we have real synthesizable performance counters exposed to pytorch then we could imagine running the exact same workload analysis we use on cosim with F1 to get results faster for workload performance charaterization!

taylor-bsg:
Probably we will not use F1, we would use the HammerBlade FPGA system..

PR #122, opened 2020-06-29
Work load evaluation toolchain used in June milestone

yodada:
This PR includes all the changes I made for the June milestone -- end to end performance and energy evaluation of Recsys training. It is a huge ... PR ... :(

Included in this PR:
 - a set of post processing scripts under `/hammerblade/scripts`.
 - `addmm`, `reduction`, and copy on device performance tuning.
 - All tensor meta data is stored on in DMEM. 
 - Now we run an ATen op twice when it is `redispatched` -- then the computation continues with CPU results. Also, after finishing `redispatching`, the user has the option to cross check results with CPU.
 - When an ATen op is `redispatched`, per op profiling data and input dimensions are printed.
 - Now the device code has an imaginary `__bsg_pod_id` and `BSG_POD_DIM`. They should be 0 and 1 respectively, unless you are doing workload eval.
 - Fixed a bug where device tensor allocator does not call wrapper functions in `HammerBladeFunctions` -- and thus not being `Trimmed` correctly when being profiled.
 - Fixed a number of bugs in `redispatching` logic

There *is* one known issue in the code -- if you want to do native profiling, the current way of dumping per ATen op profiling data may artificially slow it down. Look at `HBProfilerLog::~HBProfilerLog()` for more details.

This PR includes the changes in #114 , so I'll close that one.


hammerblade/torch/kernel/hb_tensor.hpp
```diff
@@ -129,25 +131,22 @@ class HBTensorImpl {
...
+                  + index1 * strides[1]
+                  + index2 * strides[2]];
+    }
 
-      return data[offset];
+    DT& operator()(uint32_t index0, uint32_t index1, uint32_t index2, uint32_t index3) {
+      return data[index0 * strides[0]
+                  + index1 * strides[1]
+                  + index2 * strides[2]
+                  + index3 * strides[3]];
```
vb000, line 149: `initializer_list` is a compile-time concept and is perfect to avoid code duplication here. Two things we have to add now that we are optimization phase:
- Unroll the initializer list loop.
- `hb_assert_msg`s could be disabled by default and wrap them in `ifdef`s:
```
#ifndef DEBUG
#define hb_assert_msg
#else
#define hb_assert_msg <actual definition>
```

I tried these here: https://godbolt.org/z/zXtSdX. We get exact same codegen with initializer_list without needing to duplicate code!

hammerblade/torch/kernel/hb_tiled_for.hpp
```diff
@@ -260,13 +345,25 @@ inline void hb_tiled_foreach(HBTensor<scalar_t> res,
     strides[2] = (tensor1.get_strides())[0];
     strides[3] = (tensor2.get_strides())[0];
```
vb000, line 350: We don't have to use byte strides since we only support `uint32_t` and `float`. Having to support all possible types like `uint8_t`, `uint16_t` and `uint64_t` unnecessarily adds multiplies during indexing, and even if we do that there are other parts of hb-pytorch backend that prevents us from using this feature anyway! But this a separate idea we can discuss in an issue probably.

hammerblade/torch/kernel/hb_tiled_for.hpp
```diff
@@ -446,27 +570,107 @@ inline void hb_tiled_foreach_conversion(HBTensor<scalar_dst> res,
...
+        scalar_dst* res_dp_3 = (scalar_dst*)(data[0]);
+        data[0] += strides[0];
+        data[1] += strides[1];
+
+        *res_dp_0 = functor(input_dp_0);
+        *res_dp_1 = functor(input_dp_1);
+        *res_dp_2 = functor(input_dp_2);
+        *res_dp_3 = functor(input_dp_3);
+      }
+    }
```
vb000, line 612: Normal strides also saves us from having to use this pointer arithmetic!

c10/hammerblade/HammerBladeFunctions.cpp
```diff
@@ -14,6 +14,10 @@ const int IDLE = -1;
 const int IN_USE = -42;
 std::atomic<int> hb_device_status{IDLE};
 
+// bsg_time to cycles conversion factor
+// check #87
+const uint64_t cycle_time = 10000;
```
vb000, line 19: This is 400000 (displayed at the beginning of cosim run).

vb000:
How is work distributed based on `__bsg_pod_id`? My understanding is that a chunk of batch is run by each pod automatically, even though host is oblivious to this and only knows about full batch. Correct?

vb000:
Does `redispatched` means an operator not implemented by HB is emulated on CPU?

vb000:
What does `trim` mean here?

vb000 (approved):
Lgtm! Except that we should continue to use variadic templates with `initializer_list` as the codegen looks as good as the expanded version.

yodada:
Oh this is a trim https://en.wikipedia.org/wiki/Trim_tab

Just kidding ;) The idea of the whole `trimming` things is basically "replacing _simulator_time_ with _simulated_time_"

hammerblade/torch/kernel/hb_tensor.hpp
yodada: Somehow I had this impression that using `initializer_list` results in strange code after unrolling

hammerblade/torch/kernel/hb_tiled_for.hpp
yodada: I can totally see what you are saying ;)

yeah the problem of TensorIterator is ... it's just complicated. Just realized that it actually reorder your dimensions. Like a(x, y, z), normally x is index0 and y is index1. But TensorIterator reverse it. It makes z index0 instead of index2.

yodada:
> How is work distributed based on `__bsg_pod_id`? My understanding is that a chunk of batch is run by each pod automatically, even though host is oblivious to this and only knows about full batch. Correct?

Yes and yes

yodada:
> Does `redispatched` means an operator not implemented by HB is emulated on CPU?

It really means you start the program on CPU but there is one ATen op you want to run on HB instead. From pytorch's point of view, everything happens on CPU.

The unimplemented counterpart is called `fallback`

yodada:
The TensorIterator reorder things get really complicated with reductions.

Say you want to reduce a tensor (x, y, z) along y-axis.
First, TensorIterator reverses the indices -> (z, y, x)
Next, TensorIterator pushes the reduction dim to index0 -> (y, z, x)

When I first write the reduction kernel, I assumed "pushes reduction dim to index0" means (x, y, z) -> (y, x, z)

hammerblade/torch/kernel/hb_tiled_for.hpp
vb000: But byte strides would still be in the last dimension?

yodada: Emmm I guess my point is, even without this strides in bytes thing, we still need to distinguish between "regular tensor" and "iterator tensor", because of the reordering. Well technically it won't be a correctness issue, I guess. Basically that means we treat a (a, b, c) tensor as a (c, b, a) tensor. As long as the strides and sizes are correct (they are reordered as well), we have correct results.

I mean I have no strong objection of removing the bytes ;)

PR #124, opened 2020-06-30
Bladerunner v4.1.0

yodada:
Adapt PyTorch to use Bladerunner v4.1.0

### DIFF
--- a/hammerblade/torch/pycosim.sh
+++ b/hammerblade/torch/pycosim.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/bash -x
 
-COSIM_PYTHON_EXE=$BRG_BSG_BLADERUNNER_DIR/bsg_replicant/testbenches/pytorch/test_loader
+COSIM_PYTHON_EXE=$BRG_BSG_BLADERUNNER_DIR/bsg_replicant/examples/python/test_loader
 
 if [[ ! -f $COSIM_PYTHON_EXE ]]
 then
--- a/setup_cosim_build_env.sh
+++ b/setup_cosim_build_env.sh
@@ -42,12 +42,9 @@ else
 fi
 
 # Build COSIM runtime library and simulation executable
-make -C $BRG_BSG_BLADERUNNER_DIR/bsg_replicant/testbenches/pytorch test_pytorch.log
-make -C $BRG_BSG_BLADERUNNER_DIR/bsg_replicant/testbenches/pytorch test_pytorch.debug.log
-
-# For backward compatibility.
-# Remove this with bsg_bladerunner's next version. Current is v4.0.0.
-make -C $BRG_BSG_BLADERUNNER_DIR/bsg_replicant/testbenches/python test_loader
+make -C $BRG_BSG_BLADERUNNER_DIR/bsg_replicant/examples/python test_python.log
+# this one does not work ... what should we do?
+# make -C $BRG_BSG_BLADERUNNER_DIR/bsg_replicant/examples/python test_python.debug.log
 
 export HB_KERNEL_DIR=$DIR/hammerblade/torch
 
--- a/c10/hammerblade/CMakeLists.txt
+++ b/c10/hammerblade/CMakeLists.txt
@@ -58,7 +58,7 @@ else()
   if( NOT DEFINED ENV{BSG_MANYCORE_DIR} )
     message( FATAL_ERROR "BSG_MANYCORE_DIR not set: Please set BSG_MANYCORE_DIR to bsg_f1/libraries location.")
   endif()
-  find_library(BSG_MANYCORE_RUNTIME_LIB bsg_manycore_runtime HINTS $ENV{BSG_MANYCORE_DIR})
+  find_library(BSG_MANYCORE_RUNTIME_LIB bsg_manycore_runtime HINTS $ENV{BSG_MANYCORE_DIR}/platforms/aws-vcs)
   target_link_libraries(c10_hammerblade PUBLIC ${BSG_MANYCORE_RUNTIME_LIB})
 
   # Dependency on kernel binary path
--- a/hammerblade/torch/pycosim.trace.sh
+++ b/hammerblade/torch/pycosim.trace.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/bash -x
 
-COSIM_PYTHON_EXE=$BRG_BSG_BLADERUNNER_DIR/bsg_replicant/testbenches/pytorch/test_loader
+COSIM_PYTHON_EXE=$BRG_BSG_BLADERUNNER_DIR/bsg_replicant/examples/python/test_loader
 
 if [[ ! -f $COSIM_PYTHON_EXE ]]
 then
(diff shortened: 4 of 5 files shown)


hammerblade/torch/pycosim.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/bash -x
 
-COSIM_PYTHON_EXE=$BRG_BSG_BLADERUNNER_DIR/bsg_replicant/testbenches/pytorch/test_loader
+COSIM_PYTHON_EXE=$BRG_BSG_BLADERUNNER_DIR/bsg_replicant/examples/python/test_loader
```
yodada, line 3: @vb000 I know v4.1.0 is not finalized, but my question is, are we going to use `python` instead of `pytorch`? I mean it makes sense as they are identical

setup_cosim_build_env.sh
```diff
@@ -42,12 +42,9 @@ else
...
 # Build COSIM runtime library and simulation executable
-make -C $BRG_BSG_BLADERUNNER_DIR/bsg_replicant/testbenches/pytorch test_pytorch.log
-make -C $BRG_BSG_BLADERUNNER_DIR/bsg_replicant/testbenches/pytorch test_pytorch.debug.log
-
-# For backward compatibility.
-# Remove this with bsg_bladerunner's next version. Current is v4.0.0.
-make -C $BRG_BSG_BLADERUNNER_DIR/bsg_replicant/testbenches/python test_loader
+make -C $BRG_BSG_BLADERUNNER_DIR/bsg_replicant/examples/python test_python.log
+# this one does not work ... what should we do?
+# make -C $BRG_BSG_BLADERUNNER_DIR/bsg_replicant/examples/python test_python.debug.log
```
yodada, line 47: @vb000 emmm ... I didn't figure out how to build the `debug` version ...

yodada:
This PR will be delayed until Bladerunner v4.1.0 is merged into master ;)

setup_cosim_build_env.sh
drichmond: I'll look into this. It should work

hammerblade/torch/pycosim.sh
drichmond: Did I remove something I shouldn't have?

yodada: The folder is still there, but the Makefile is not working. I tried modifying it according to the one under python but I didn't get it to work. I'm sure I missed something there ...

setup_cosim_build_env.sh
yodada: Thanks! ;)

hammerblade/torch/pycosim.sh
drichmond: OK. So I probably skipped updating it since I didn't realize it was critical.

vb000: Python is okay I think, because all we need is a way to start Python interpreter.

vb000:
I've been working with latest bladerunner for a while, and as a result #126 includes these updates...

It also includes one more change to reflect the fact that cudalite runtime api (headers) remain in `bsg_replicant/libraries` whereas libraries are built in `bsg_replicant/libraries/platforms/aws-vcs`. So I added two variables there, called `BSG_MANYCORE_INCLUDE` and `BSG_MANYCORE_LDPATH`.

yodada:
> I've been working with latest bladerunner for a while, and as a result #126 includes these updates...
> 
> It also includes one more change to reflect the fact that cudalite runtime api (headers) remain in `bsg_replicant/libraries` whereas libraries are built in `bsg_replicant/libraries/platforms/aws-vcs`. So I added two variables there, called `BSG_MANYCORE_INCLUDE` and `BSG_MANYCORE_LDPATH`.

Sweeeeet! I should have chance to look at #126 today ;)

drichmond:
Actually, the library is built in BSG_PLATFORM_PATH. 

Dustin Richmond (He/Him)
Postdoctoral Research Associate
University of Washington
www.dustinrichmond.com

vb000:
Dustin, the variables I mentioned are from hb-pytorch's perspective used in hb-pytorch repo. To basically provide info about where to find includes & where to find the library.

drichmond:
Yes, but you can't guarantee that it will always be in aws-vcs. For example, it might be in dpi-verilator. You should use BSG_PLATFORM_PATH.

vb000:
The path depends on which platform pytorch is compiled for and each platform would set a different value in `BSG_MANYCORE_LDPATH`. We could set it to `BSG_PLATFORM_PATH` in initial configure script if all platforms are drop-in replacements to others. But having a variable to know where to find `.so` file is important to minimize dependencies across repos. For example when the name `BSG_PLATFORM_PATH` changes, we wouldn't have to worry about it's uses in all CMake files in pytorch but can just update `BSG_MANYCORE_LDPATH` accordingly.

drichmond:
It seems that your main concern is that the variable BSG_PLATFORM_PATH might change. I don't intend to change BSG_PLATFORM_PATH. 

But your concern isn't unfounded. We all have intentions until some other problem comes up and someone (me) refactors the makefiles. 

But this discussion is somewhat orthogonal to my original point: Don't hardcode aws-vcs.

vb000:
Well, we wouldn't want platform set in bsg_bladerunner affect what backend is used in pytorch, right? Or, am I not understanding BSG_PLATFORM_PATH correctly?

drichmond:
That is up to you. I understand if you don't.

BSG_PLATFORM_PATH is designed to be independent of bladerunner.

The link/compile fragments that use BSG_PLATFORM_PATH in replicant are designed to be independent of bladerunner (that's why environment.mk is ONLY included in files called Makefile)

As I said though, in the end, all I'm trying to avoid is having you hardcode aws-vcs. The runtime library for each platform are designed to be swappable (e.g. you could use LD_LIBRARY_PATH to switch between them, but please don't :P. There are better ways)

Likewise in verilator, the libmachine.so (that defines the machine) is also designed to be swappable. Don't want to use 4x4_fast_n_fake and you don't want to recompile? Switch the path to libmachine.so.

vb000:
I feel platform abstraction is strong enough that different apps can choose different platforms at compile-time (or runtime if APIs are consistent and platforms are swappable). In that sense, BSG_PLATFORM_PATH restricts multiple apps to use same platform. 

I agree that hardcoding path is a downside with the current approach. That's the reason I hardcoded it in only one place and it becomes BSG_MANYCORE_LDPATH from there on. And that's also the reason why I requested for standard paths: it'd be fantastic if we could "install" headers and `.so` files in `<custom-prefix>/include` and `<custom-prefix>/lib`. If that's possible, pytorch could just build `bsg_bladerunner` and install headers, libraries and executables at known locations.

vb000:
>And that's also the reason why I requested for standard paths: it'd be fantastic if we could "install" headers and `.so` files in `<custom-prefix>/include` and `<custom-prefix>/lib`. If that's possible, pytorch could just build `bsg_bladerunner` and install headers, libraries and executables at known locations.

Basically I'm suggesting GNU autotools/CMake style build, install flows in `bsg_bladerunner`. We could even have configure step to configure for each platform.

PR #125, opened 2020-06-30
Strides of TensorIterator tensors are no longer measured in bytes

yodada:
Now tensors retrieved from TensorIterators also have their strides measured in `num elements`. It is the kernel author's responsibility to ensure the type safety.


zzy82518996:
Cool! Is this means that I can directly offload the kernel and use tile_range function inside the kernel ?

vb000 (approved):
Great!

yodada:
> Cool! Is this means that I can directly offload the kernel and use tile_range function inside the kernel ?

Well yes and no ;)

One thing we've been talking about is TensorIterator orders the dimensions ... which can be annoying, but will not result in wrong answers -- if you have a regular tensor (which is not part of an iterator) with shape a x b x c, if you use `hb_tiled_foreach`, it will loop through it as if it has shape c x b x a ...

Since `hb_tiled_foreach` is designed for elementwise ops, it yields the correct results, but may induce worse performance due to altered memory access patterns.

PR #127, opened 2020-07-02
Non-blocking loads in elementwise kernels

vb000:
- Updated cosim build to run with hardfloat and latest updates to bsg_replicant.
- Added following definitions to help us write code generating non-blocking loads:
   - `__remote`: pointer can be annotated as remote. For example:

       ```
        __remote float* foo;
       ```
     would essentially declare `foo` as `__remote float*` type.
   - `NOALIAS`: A wrapper for `restrict`, `__restrict` or `__restrict__`.  A pointer in a function arguments' list can be qualified as `NOALIAS` to hint compiler that this pointer will not be aliased in the scope of the function. In the following example:
https://github.com/cornell-brg/hb-pytorch/blob/a7d504c6177921f62ecad9e854f1e3bcd1b904db/hammerblade/torch/kernel/hb_tiled_for.hpp#L100-L103
       the *data pointed by `res_ptr`* (not the pointer itself) is qualified with `NOALIAS`.
   - Unrolling pragma is slightly different for GCC and Clang. So, this defines a wrapper macro `UNROLL` to automatically select the right pragma:
https://github.com/cornell-brg/hb-pytorch/blob/a7d504c6177921f62ecad9e854f1e3bcd1b904db/hammerblade/torch/kernel/hb_common.hpp#L26-L31
      Using this, a loop can be unrolled like this:
     ```c++
      UNROLL(16) for(size_t idx = start; idx < end; idx++) { 
         ...
      }
      ```
- Refactored [`hb_tiled_foreach`](https://github.com/cornell-brg/hb-pytorch/blob/a7d504c6177921f62ecad9e854f1e3bcd1b904db/hammerblade/torch/kernel/hb_tiled_for.hpp#L80-L96) to effectively employ `NOALIAS`.  This implementation does a static dispatch to helper functions with `NOALIAS` annotations based on the number of operands.
- Some results of 10k + 10k tensor-add on 4x4_fast_n_fake:
   1. GCC without `NOALIAS` or `UNROLL`:
	```
	Abs Total Cycles                 IPC
	=====================================
	           15775              0.3854
	=====================================
	
	stall_depend_dram_load 102599
	```
  2. GCC with `NOALIAS` and `UNROLL`:
	```
	Abs Total Cycles                 IPC
	====================================
	           10559              0.4921
	====================================
	
	stall_depend_dram_load 15839
	```
   3. GCC without `NOALIAS` and with `UNROLL`:
	```
	Abs Total Cycles                 IPC
	=====================================
	           17894              0.3025
	=====================================
	
	stall_depend_dram_load 108827
	```
	
  4. Clang with `NOALIAS` & `UNROLL`:
	```
	Abs Total Cycles                 IPC
	====================================
	           19815              0.3082
	====================================
	
	stall_depend_dram_load 81386
	```
  Analysis of codegen is more important here and those numbers just show which configuration makes a +ve difference and which one make -ve difference. More importantly, in the disassembly (and also in the IR in case of Clang), `NOALIAS` is causing the the loads to reorder, and with `NOALIAS`, both GCC and Clang are generating non-blocking loads. Other than non-blocking loads, Clang codegen is not generating `fma`s, but that doesn't fully explain the poor performance. Clang didn't fully utilize non-blocking loads even though there's is concrete evidence of Clang annotating pointers with `noalias` in the IR, and considering remote load latency to be 20 cycles. With `NOALIAS`, GCC is also hiding memory latency with non-blocking loads to decent extent, as is evident in ii. More latency hiding is possible using more FP registers, I'll try to figure out the reason for Clang's poor performance and try to achieve perfect schedule.


taylor-bsg:
Looking forward to learning more!

hammerblade/torch/kernel/hb_tiled_for.hpp
```diff
@@ -74,218 +74,121 @@ inline void calc_range(hb_range* range, size_t numel) {
...
-    data[1] += strides[1] * start;
-    data[2] += strides[2] * start;
-    for (size_t idx = start; idx < end; idx++) {
-      *data[0] = functor(*data[1], *data[2]);
-      data[0] += strides[0];
-      data[1] += strides[1];
-      data[2] += strides[2];
+    UNROLL(16) for(size_t idx = start; idx < end; idx++) {
+      res_ptr[idx * res.get_strides()[0]] =
+        functor(tensor_data_ptr0[idx * tensor_arg0.get_strides()[0]]);
```
yodada, line 130: How does the assembly look like? does it have the multiplication in the inner loop? I guess no?

yodada (approved):
Looks awesome! Just want to make sure there is no multiplication in the inner loop ;)

hammerblade/torch/kernel/hb_tiled_for.hpp
vb000: No, they are compiled to a series of adds... felt `data[idx * stride]` format is easier to read.

PR #133, opened 2020-07-16
BatchNorm2d Implementation

vb000:


torch/nn/modules/module.py
```diff
@@ -297,7 +297,16 @@ def hammerblade(self):
         Returns:
             Module: self
         """
-        return self._apply(lambda t: t.hammerblade())
+        def apply_hb(t):
+            if isinstance(self, torch.nn.BatchNorm2d) and \
+                    (t.dtype == torch.int64):
+                # BatchNorm2d's batch counter is a long tensor. Since it is metadata
```
yodada, line 303: is this the 64bit integer add we saw?

setup_cosim_build_env.sh
```diff
@@ -44,8 +44,9 @@ else
 fi
 
 # Build COSIM runtime library and simulation executable
+export BSG_MACHINE=4x4_fast_n_fake
```
yodada, line 47: this will override the settings in `bsg_replicant/machine.mk` right?

yodada (approved):
Awesome! Thanks a lot for implementing this

torch/nn/modules/module.py
vb000: Yes, later found out this not the core tensor in the kernel computation. This is just the metadata of `torch.nn.BatchNorm2d` layer. Meaning `torch.nn.functional.batch_norm` doesn't need this.

setup_cosim_build_env.sh
vb000: Yes, just wanted to be able to set the machine in pytorch, independent of what's set in bsg_replicant.

Issue #121, opened 2020-06-25
Addmm Experiment


vb000:
As discussed in the meeting, I was looking at ways to use NB loads in Addmm fo the deadline. I started with a simpler version and to my surprise, this simpler version without blocking/scratchpad is performing slightly better than the blocking version in the master! Albeit, I only tried this on an 8x8 micro-benchmark.

It's as simple as this:
https://github.com/cornell-brg/hb-pytorch/blob/2f9f01bf711c6343cdd849d56a3050280c25a6eb/hammerblade/torch/kernel/kernel_addmm.cpp#L34-L50

@yodada Is the addmm on master parallelized? If so and this still is performing slightly better or on-par with the version on master, I think it makes sense to implement blocking & nb-load optimizations on this baseline algorithm.

Two main ideas in this implementation:
1. 2D partition of work among tiles as opposed to linear `hb_tiled_for`
2. The `k` dimension should be in the outer loop. This makes sure that when a tile loads a column or row, it make sure full use of the data and don't have to reload it again.

taylor-bsg:
Interesting!

If we are changing what we do based on a performance measurement, let’s be
sure to specify: 1) the machine model 2) specifying how many tiles were
used, 3) a hypothesis about why A runs better than B which should be in
terms of work performed, or reuse at a particular level of the memory
hierarchy 4) comparative data from either assembly dumping or execution
statistics that confirms the hypothesis and 5) an explanation of why the
data set being used is representative (or if it is not clear, do a sweep
among sizes that do not fit at each relevant level of the memory hierarchy).

M

PR #136, opened 2020-07-20
Parallel Convolution

vb000:
- Parallel forward and backward convolutions:
  - Forward: output tensor is distributed among tiles.
  - Backward:
     - Input gradients: batches and input channels are distributed among tiles.
     - Weight gradients: Input and output channels are distribute among tiles.
  - Also includes a couple scheduling optimizations.
  - ~90x speedup on ResNet workload on the chip.
- Added multi-indexed-tiled-for (`hb_tiled_for(<lambda>, index1, index2, ...)`) to simplify loop scheduling optimizations. This also eliminates the need to manually compute indices by `/` and `%` on monolithic index.
- Added `hb_spatial_for` utility function to be able to do 2D spatial distribution among tiles, as opposed to distributing single monolithic index among tiles. I had found this useful when I experimented it on addmm (https://github.com/cornell-brg/hb-pytorch/issues/121), so tried this on Conv2d as well. It turns out that images dimensions are not large enough to make full use of manycore array for Conv2d. But we should definitely try this on addmm.

Merge after #135

### DIFF
--- a/hammerblade/torch/kernel/hb_spatial_for.hpp
+++ b/hammerblade/torch/kernel/hb_spatial_for.hpp
@@ -1,0 +1,33 @@
+// =========================================================
+// HB Spatial For
+// 
+// 07/19/2020 Bandhav Veluri
+// =========================================================
+
+// ---------------------------------------------------------
+// HB 2D spatial for
+// ---------------------------------------------------------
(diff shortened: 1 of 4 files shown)


yodada (approved):
LGTM!

PR #137, opened 2020-07-22
Foreach unroll

kr397:
Implemented foreach for element-wise kernels with unrolling factor N as a template argument. Changed every element-wise kernel to call hb_tiled_foreach_unroll with the optimum unrolling factor (based on COSIM evaluations conducted on the 4x4 Bladerunner). Results of the evaluations can be found here: 
[https://docs.google.com/spreadsheets/d/1cfCi62IBBbtfcwqjdckJ_QPopBFZ9bYTNk54ioiZeHE/edit?usp=sharing](url)


cbatten:
great start! I think we want to do performance tuning on the 8x16 timing configuration though?

kr397:
Thanks! Yeah it should eventually be for 8x16, I did it for 4x4 first since it's faster. I'll update the unrolling factors!

aten/src/ATen/native/hammerblade/Vvadd.cpp
```diff
@@ -11,6 +11,7 @@ Tensor vvadd_hb(const Tensor& self, const Tensor& other) {
 
   // Tutorial TODO:
   // Call HB device kernel tensorlib_vvadd
+  hb_offload_kernel(result, self, other, "tensorlib_vvadd");
```
yodada, line 14: I guess this got into the PR accidentally? We might want to keep vvadd unimplemented for future ppl to work on

hammerblade/torch/kernel/hb_reduction.hpp
```diff
@@ -69,9 +69,9 @@ inline void binary_reduction_simple(HBTensor<scalar_t> out,
                                     F1 reduce, F2 project) {
   hb_assert_msg(out.numel() == 1, "reduction_simple only handles trivial case");
 
-  __remote scalar_t* data[2];
-  data[0] = (__remote scalar_t*)out.data_ptr();
-  data[1] = (__remote scalar_t*)in.data_ptr();
+  scalar_t* data[2];
```
yodada: Hmmm we also dont want to revert what Bandhav did. So these need to be __remote still

hammerblade/torch/kernel/kernel_conv.cpp
```diff
@@ -129,7 +129,7 @@ extern "C" {
     bsg_cuda_print_stat_kernel_start();
 
     // init input grads
-    hb_tiled_foreach([]() {return 0.0;}, x);
+    hb_tiled_foreach(x, []() {return 0.0;});
```
yodada: Ok for this syntax we need @vb000 's input. So @vb000 , do you have to put the lambda at the very beginning for your code to work?

vb000: No, I was trying to make this work with variadic arguments to avoid repeating the implementation, in which case, tensor arguments have to be in the end. For example,

```
hb_tiled_foreach(<lambda>, T... tensor_args)
```

is supposed to handle all variants. `tensor_args` is a parameter pack here, and to achieve this with `hb_tiled_foreach`, we need to iterate over each argument in the pack and apply a computation. But in C++11, there doesn't seem to be a way to do compile time loops on parameter packs. So, I couldn't get this done. But I moved tensor arguments to the end, in case any meta-programming wizard wants to tackle this ;). We can move lambda to the end if that's convenient for any of your implementations.

aten/src/ATen/native/hammerblade/Vvadd.cpp
kr397: Oh yes sorry about that, I just removed all the things I had added during the tutorial

hammerblade/torch/kernel/hb_reduction.hpp
kr397: Yes okay reverted that

hammerblade/torch/kernel/kernel_conv.cpp
kr397: Yes, I tried using variadic arguments for that too, but it seemed like there would still have to be different implementations for different operations on some level so I worked with the overloaded function itself. However, my implementation is indifferent to the position of the lambda. I think @cbatten preferred having the lambda at the end.

cbatten:
I would definitely like to have the lambda at the end. I think it is much more natural, avoids the tensors getting "lost" at the end of a multi-line lambda, and follows the TBB parallel_for convention ...

hammerblade/torch/kernel/hb_reduction.hpp
```diff
@@ -103,7 +103,7 @@ inline void binary_reduction_simple(HBTensor<scalar_t> out,
     // iterating over all elementes
     //-----------------------------
     hb_tiled_for(in.numel(), [&](size_t idx) {
-      __remote scalar_t* in_dp = (__remote scalar_t*)(data[1] + offset_calc(idx, in));
+      __remote scalar_t* in_dp = (scalar_t*)(data[1] + offset_calc(idx, in));
```
yodada: I guess we still need the __remote here?

hammerblade/torch/riscv.mk
```diff
@@ -96,13 +94,8 @@ kernel.riscv: $(KERNEL_OBJS)
...
 		-L. -l:$(BSG_MANYCORE_LIB) -o $@ $(filter-out -nostdlib,$(RISCV_LINK_OPTS))
 
-$(KERNEL_DIS): %.dis: %.o
-	$(RISCV_BIN_DIR)/riscv32-unknown-elf-dramfs-objdump -M numeric --disassemble-all -S $< > $@
-
-kernel.dis: kernel.riscv
-	$(RISCV_BIN_DIR)/riscv32-unknown-elf-dramfs-objdump -M numeric --disassemble-all -S $< > $@
-
-all.dis: kernel.dis $(KERNEL_DIS)
+%.dis: %.o
```
yodada: I think in your local branch, you reverted the commit that leads to empty /torch/riscv/. So you would need to "revert the revert"

hammerblade/torch/kernel/hb_tiled_for.hpp
```diff
@@ -76,121 +76,218 @@ inline void calc_range(hb_range* range, size_t numel,
 }
 
 // =========================================================
-// Tiled Pointwise for
+// Tile Pointwise for -- Ternary
 // =========================================================
```
yodada, line 81: So ... do we have a conclusion here? What should we do?

yodada (commented):
I also saw you included a change to `third_party/fbgemm`. Somehow we are keep updating this submodule. Can you see if your fbgemm is on commit `7d67304`?

hammerblade/torch/riscv.mk
kr397: Oh yes, there was a problem with merge which I'd fixed but not pushed; done that now

hammerblade/torch/kernel/hb_reduction.hpp
kr397: Yes, sorry about that

hammerblade/torch/kernel/hb_tiled_for.hpp
kr397: Do you mean the order of arguments?

yodada: ahh yes

hammerblade/torch/kernel/hb_tiled_for.hpp
```diff
@@ -233,22 +330,22 @@ inline void hb_tiled_foreach_conversion(HBTensor<scalar_dst> res,
     if (end - start > 4) {
       for (; idx < end - 4; idx += 4) {
         scalar_src input_dp_0 = *(input_data);
-        __remote scalar_dst* res_dp_0 = (res_data);
+        scalar_dst* res_dp_0 = (res_data);
```
yodada: So here I just want to make sure we are not just reverting what @vb000 have. And if we do, we need good reasons.

Are those `remote` keywords removed because of conflict resolving or because we are trying to put functor at the end?

PR #138, opened 2020-07-23
Conv2d backward scheduling optimizations and micro benchmarking

vb000:
- Loop scheduling optimizations for Conv2d backward.
- A framework for micro-benchmarking. This generates an output like this:
```
     Layer |          Layer parameters |         Inputs shape |   CPU Time (ms) |    HB Time (ms) |  HB FLOPs/cycle
    Conv2d |       [[8, 3, 3, 3], [8]] |         [1, 3, 8, 8] |            0.07 |           17.31 |          0.0005
    Conv2d |       [[8, 3, 3, 3], [8]] |       [1, 3, 32, 32] |            0.18 |           17.40 |          0.0118
    Conv2d |       [[8, 3, 3, 3], [8]] |       [1, 3, 64, 64] |            0.47 |           20.23 |          0.0435
    Conv2d |       [[8, 3, 3, 3], [8]] |       [8, 3, 32, 32] |            0.97 |           23.74 |          0.0694
    Conv2d |       [[8, 3, 3, 3], [8]] |       [8, 3, 64, 64] |            2.91 |           54.98 |          0.1279
    Conv2d |       [[8, 3, 3, 3], [8]] |      [16, 3, 32, 32] |            1.62 |           32.48 |          0.1014
    Conv2d |       [[8, 3, 3, 3], [8]] |      [16, 3, 64, 64] |            5.09 |           98.05 |          0.1435
    Conv2d |       [[8, 3, 3, 3], [8]] |      [32, 3, 32, 32] |            3.34 |           52.01 |          0.1266
    Conv2d |       [[8, 3, 3, 3], [8]] |      [32, 3, 64, 64] |           10.49 |          185.22 |          0.1519
    Conv2d |       [[8, 3, 3, 3], [8]] |      [64, 3, 32, 32] |            5.48 |           92.84 |          0.1419
    Conv2d |       [[8, 3, 3, 3], [8]] |      [64, 3, 64, 64] |           20.39 |          407.14 |          0.1382
    Conv2d |     [[16, 8, 3, 3], [16]] |         [1, 8, 8, 8] |            0.08 |           17.39 |          0.0025
    Conv2d |     [[16, 8, 3, 3], [16]] |       [1, 8, 32, 32] |            0.44 |           21.01 |          0.0511
    Conv2d |     [[16, 8, 3, 3], [16]] |       [1, 8, 64, 64] |            1.58 |           37.90 |          0.1209
[...]

PR #139, opened 2020-07-24
Install disassembly with pytorch build

vb000:
Builds and installs disassembly files along with kernel binary. This makes opening disassembly very convenient, with the added benefit of making sure we are running and looking at the exact same version of machine code.

### DIFF
--- a/c10/hammerblade/CMakeLists.txt
+++ b/c10/hammerblade/CMakeLists.txt
@@ -75,12 +75,12 @@ else()
   # Kernel binary
   add_custom_target(
     kernel_riscv_exe
+    COMMAND rm -rf riscv
     COMMAND mkdir -p riscv
     COMMAND make -C riscv -f $ENV{HB_KERNEL_DIR}/riscv.mk ${KERNEL_MAKE_FLAGS} kernel.riscv -j8
-    COMMAND cp riscv/kernel.riscv .
-    COMMAND rm -rf riscv
+    COMMAND make -C riscv -f $ENV{HB_KERNEL_DIR}/riscv.mk ${KERNEL_MAKE_FLAGS} all.dis -j8
     DEPENDS $ENV{HB_KERNEL_DIR}/riscv.mk
-    BYPRODUCTS kernel.riscv
+    BYPRODUCTS riscv/
     )
 
   add_dependencies(c10_hammerblade kernel_riscv_exe)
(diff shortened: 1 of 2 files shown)


yodada:
These changes are included in #138 's change set I believe? Do we need to merge them separately?

vb000:
Oh, yeah, I had rebased that branch on it as I was using these changes there... we can just merge that if changes look good to you.

yodada:
Looks all good!

c10/hammerblade/CMakeLists.txt
```diff
@@ -120,8 +120,11 @@ endforeach()
 install(FILES ${CMAKE_BINARY_DIR}/c10/hammerblade/impl/hammerblade_cmake_macros.h
   DESTINATION include/c10/hammerblade/impl)
 
+file(GLOB KERNEL_FILES ${CMAKE_CURRENT_BINARY_DIR}/riscv/kernel.riscv
+                       ${CMAKE_CURRENT_BINARY_DIR}/riscv/*.dis)
+
 if(NOT USE_HB_EMUL)
-  install(FILES ${CMAKE_CURRENT_BINARY_DIR}/kernel.riscv DESTINATION riscv)
+  install(FILES ${KERNEL_FILES} DESTINATION riscv)
```
yodada, line 127: @vb000  I think after this change, `torch/riscv` is missing when building from a clean clone. somehow these files are not "installed" correctly

yodada: @zzy82518996 you saw similar issues I believe?

zzy82518996: Yes, I also came across the same issue.

vb000: Not sure why this is the case with fresh builds. I'll look into it...

yodada: not too sure as well, but it seems to me this `install` command didnt do anything. I was wondering if the setting the entire path `riscv/` as `BYPRODUCTS` is not supported by cmake?

Btw, I dont have `ninja` installed, is that a must have for this?

vb000: No, I think. The byproducts thing only affects builds and that too only on the `clean` target because we rebuild every single time because of the cross compilation issues. Could you confirm if they are correctly built in `build/c10/hammerblade/riscv/`?

yodada: Yes, everything is there in `build/c10/hammerblade/riscv/`, including all the` *.dis`
but none of them are copied over to `torch/riscv/`

vb000: You know what to do... just copy them to torch/riscv ;). JK, I'll try to figure out the issue.

I tried removing my CMakeCache, but still it seems to install fine! So, I'll try a fresh build...

yodada: yeah we are manual copying them as a work around of course! ;)

can you share your cmake version maybe? we are using `cmake version 3.13.3`

vb000: `cmake3 version 3.17.1`

yodada: maybe I need a version bump

which version are you using? @zzy82518996

zzy82518996: my cmake version is 3.18.0
my cmake3 version is 3.17.3

zzy82518996: this is inside my python virtual environment.

PR #140, opened 2020-07-24
[setup_cosim] adjust setup_cosim script so we can share COSIM install…

yodada:
- Don't rebuild python test_loader if running on one of the BRG servers -- we share a single global installation of COSIM. `BSG_MACHINE` and `BSG_MACHINE_PATH` will be defined in `setup-hb.sh` which should be sourced when using BRG servers.
 - Bump fbgemm version so it wont keep showing up ...


vb000 (approved):
Interesting!

PR #144, opened 2020-08-04
Fix: torch/riscv/ is not populated

yodada:
This PR fixed an issue introduced by #139 
 - Basically the problem was, `GLOB` runs before build stage, so the files we are looking for, for example, `kernel.riscv` and `tensorlib_add.dis`, are not generated yet. Thus, cmake has no idea what to install.
 - This issue is fixed by using `install(DIRECTORY` instead of `install(FILES`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -258,6 +258,7 @@ stats/
 *.cosim
 *.vpd
 *.riscv
+*.dis
 *.cosim.daidir/
 *.cosim.tmp/
 ucli.key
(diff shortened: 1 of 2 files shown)


zzy82518996 (commented):
Cool ! This really helps !

vb000 (approved):
Oh got it! Since the directory is already present during an incremental build, the issue didn't show up?

yodada:
Yeah I think so. If you clean the build dir then the problem pops up.

PR #147, opened 2020-08-06
uniform kernel

Janicewei:
It fills the tensor with numbers sampled from the continuous uniform distribution. Used in RNN.


aten/src/ATen/native/Pool.h
```diff
@@ -15,7 +15,7 @@ template <typename dest_t, typename src_t>
 static inline dest_t
 safe_downcast(src_t v)
 {
-  TORCH_CHECK(std::numeric_limits<dest_t>::min() <= v && v <= std::numeric_limits<dest_t>::max(),
+  TORCH_CHECK(std::numeric_limits<dest_t>::lowest() <= v && v <= std::numeric_limits<dest_t>::max(),
```
yodada, line 18: I think I made this change? basically the idea is `std::numeric_limits<float>::min()` gives you the smallest positive number

https://en.cppreference.com/w/cpp/types/numeric_limits/lowest

aten/src/ATen/native/hammerblade/Uniform_.cpp
```diff
@@ -0,0 +1,32 @@
+#include <ATen/Dispatch.h>
+#include <ATen/hammerblade/HammerBladeContext.h>
+#include <ATen/HammerBladeGenerator.h>
+#include <ATen/native/hammerblade/Offload.h>
+#include <ATen/native/Distributions.h>
+
+namespace at { namespace native {
+
+Tensor hammerblade_common_seed_to_tensor_(uint32_t seed = 42) {
```
yodada, line 9: There is another copy of this function in https://github.com/cornell-brg/hb-pytorch/blob/ef1429afc0ff523533520b23d25277c10b253fc1/aten/src/ATen/native/hammerblade/Distributions.cpp#L9-L23

Should we move it to a header and make it inline? (no need to change for this PR

aten/src/ATen/native/native_functions.yaml
```diff
@@ -2683,12 +2683,14 @@
...
     CPU: _tanh__cpu
     CUDA: _tanh__cuda
+    HammerBlade: _tanh__hb
 
 - func: tanh.out(Tensor self, *, Tensor(a!) out) -> Tensor(a!)
   supports_named_tensor: True
   dispatch:
     CPU: _tanh_out_cpu
     CUDA: _tanh_out_cuda
+    HammerBlade: _tanh_out_hb
```
yodada: oops. `_tanh` defined in `uniform` PR ;)

hammerblade/torch/tests/test_uniform_.py
```diff
@@ -0,0 +1,33 @@
...
+def test_torch_uniform1():
+    x = torch.rand(10, device="hammerblade")
+    out = x.cpu()
+    i = 0
+    assert (x.device == torch.device("hammerblade"))
+    while(i < x.numel()):
+        assert (out[i] < 1)
+        i += 1
+
+def test_torch_uniform2():
```
yodada, line 34: It would be also nice to test for if 
 - these numbers are > 0
 - they all not all the same

yodada (commented):
If you could remove `tanh` from `aten/src/ATen/native/native_functions.yaml`, and add another simple test that check if all values are the same, this PR would be great.

aten/src/ATen/native/UnaryOps.cpp
```diff
@@ -322,7 +322,7 @@ IMPLEMENT_UNARY_OP_VEC_CUDA(erfinv)
 IMPLEMENT_UNARY_OP_VEC_HB(exp)
 IMPLEMENT_UNARY_OP_VEC(reciprocal)
 IMPLEMENT_UNARY_OP_VEC(tan)
-IMPLEMENT_UNARY_OP_VEC(tanh)
+IMPLEMENT_UNARY_OP_VEC_HB(tanh)
```
yodada: This is also `tanh` ;)

yodada (approved):
Awesome!

PR #148, opened 2020-08-07
tanh kernel 

Janicewei:


hammerblade/torch/kernel/kernel_tanh.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+__attribute__ ((noinline))
+int tensorlib_tanh( hb_tensor_t* t0_p, hb_tensor_t* t1_p)
+{
+  auto res = HBTensor<float>(t0_p);
+  auto input = HBTensor<float>(t1_p);
+
+  bsg_cuda_print_stat_kernel_start();
+  hb_tiled_foreach(
+    [&](float a) {
```
yodada: here you can replace `[&]` with `[]` since we are not using any value that needs to be captured

not critical

yodada (commented):
Just one more thing before I approve this PR

make sure your `third_party/fbgemm` is on `7d67304`

hammerblade/torch/kernel/kernel_tanh.cpp
Janicewei: How to have fbgemm on 7b67304?

yodada: you can go to that path `third_party/fbgemm` and then `git checkout 7b67304`

yodada (approved):
Cool! just make sure we are not keep updating fbgemm then we are good!

Janicewei:
> Cool! just make sure we are not keep updating fbgemm then we are good!

Got you!

PR #149, opened 2020-08-07
simple _cat kernel, dim 0

Janicewei:


hammerblade/torch/riscv.mk
```diff
@@ -81,8 +81,6 @@ KERNEL_CSRCS   := $(notdir $(wildcard $(KERNEL_DIR)/*.c))
 KERNEL_CPPSRCS := $(notdir $(wildcard $(KERNEL_DIR)/*.cpp))
 KERNEL_OBJS    := $(patsubst %.c,%.o,$(KERNEL_CSRCS)) \
                   $(patsubst %.cpp,%.o,$(KERNEL_CPPSRCS))
-KERNEL_DIS     := $(patsubst %.c,%.dis,$(KERNEL_CSRCS)) \
-                  $(patsubst %.cpp,%.dis,$(KERNEL_CPPSRCS))
```
yodada: So here is the merge conflict -- I think you reverted `fb4a35c34ba6c`

You would need to revert the revert commit ;)

hammerblade/torch/tests/test_fallback.py
```diff
@@ -41,7 +41,6 @@ def test_native_1():
     torch.hammerblade.profiler.disable()
     assert torch.allclose(hb.cpu(), cpu)
 
-@pytest.mark.xfail
 def test_native_1F():
```
yodada: Actually, can you delete this test case? It is meant to test the `fallback` mechanism and make sure it does not do crazy things when not enabled.

hammerblade/torch/kernel/kernel_cat.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+  bsg_cuda_print_stat_kernel_start();
+  for (size_t i = 0; i < length; i++) {
+	HBTensor<float> tensor(tensors_p[i]);
+	for (size_t j = 0; j < tensor.numel(); j++) {
+	  result(index) = tensor(j);
+	  index = index + 1;
+	}
+  }
+
```
yodada, line 50: Hmmm, this is a little scary -- basically you have every tile do the same work. If you are not ready to parallelize the work, you should put `if (__bsg_id == 0)`, so tile 0 will do the work while others staying idle

aten/src/ATen/native/hammerblade/Cat.cpp
```diff
@@ -0,0 +1,193 @@
...
+  scalars.push_back(create_device_scalar(dim_i32));
+
+  // XXX: this is definitely wrong
+  Tensor result = at::empty({0}, tensors[0].options());
+  bool allSkipped = true;
+  bool allContiguous = true;
+  Tensor notSkippedTensor;
+
+  // Inputs cannot alias the output tensor
+  for (size_t i = 0; i < tensors.size(); i++) {
```
yodada: This piece of code looks strange. `result` is just allocated above `Tensor result = at::empty({0}, tensors[0].options());`, so it is not possible for it to alias with anybody.

Janicewei: updated Cat.cpp. I implemented host code based on functions in TensorShape.cpp

hammerblade/torch/tests/test_cat.py
```diff
@@ -0,0 +1,24 @@
...
+    x = torch.ones(10)
+    _test_torch_cat(x)
+
+def test_cat_2():
+    x = torch.randn(3, 4)
+    _test_torch_cat(x)
+
+def test_cat_3():
+    x = torch.randn(3, 4, 5)
+    _test_torch_cat(x)
```
yodada, line 47: More tests will make this even better ;)

For instance, I think you would want to test the case where you cat tensors with different length and empty tensors. Say, [0,1], [0,1,2,3,4,5], [] -> [0,1,0,1,2,3,4,5]

hammerblade/torch/kernel/kernel_cat.cpp
```diff
@@ -0,0 +1,44 @@
...
+  HBTensor<float> result(result_p);
+  uint32_t length = *length_p;
+  int32_t dim = *dim_p;
+  HBTensor<float> tensor(tensors_p[0]);
+  int32_t size = tensor.numel();
+  bsg_cuda_print_stat_kernel_start();
+
+  hb_tiled_for(result.numel(), [&] (size_t i) {
+    size_t j = i / size;
+    size_t index = i % size;
```
yodada: Yeah here is an assumption that isn't always true.

But as a first step, it is ok to fail on the case where input tensors have different length. However, it cannot fail silently -- you would want to check for it and throw an error

c10/hammerblade/CMakeLists.txt
```diff
@@ -75,12 +75,12 @@ else()
   # Kernel binary
```
yodada: Emmm this file shouldn't be touched. I'll help you revert the changes. Dont worry about it

hammerblade/torch/kernel/kernel_cat.cpp
```diff
@@ -0,0 +1,53 @@
...
+  bsg_cuda_print_stat_kernel_start();
+
+  hb_tiled_for(result.numel(), [&] (int32_t i) {
+    int32_t j = 0;
+    int32_t index = 0;
+    HBTensor<float> init(tensors_p[0]);
+    int32_t size = init.numel();
+    while (i >= size) {
+      index = i - size;
+      HBTensor<float> tensor(tensors_p[++j]);
```
yodada: this definitely is valid code, but personally I would avoid this. Doing `++j` inside indexing makes code hard to parse.

yodada (approved):
ok! Looks good! Let me fix the makefile before we merge.

hammerblade/torch/kernel/kernel_cat.cpp
Janicewei: Okay! changed

yodada:
Reverted `4f22ea6` -- now we are consistent with master branch. This PR is ready to merge once all checks are done.

@Janicewei have you tried on cosim? ideally we should give it a try before we merge

Janicewei:
> Reverted `4f22ea6` -- now we are consistent with master branch. This PR is ready to merge once all checks are done.
> 
> @Janicewei have you tried on cosim? ideally we should give it a try before we merge

give me a few minutes

Janicewei:
> Reverted `4f22ea6` -- now we are consistent with master branch. This PR is ready to merge once all checks are done.
> 
> @Janicewei have you tried on cosim? ideally we should give it a try before we merge

I ran `pycosim -m pytest hb-pytorch/hammerblade/torch/tests/test_cat.py -vs` and tests passed

yodada:
> > Reverted `4f22ea6` -- now we are consistent with master branch. This PR is ready to merge once all checks are done.
> > @Janicewei have you tried on cosim? ideally we should give it a try before we merge
> 
> I ran `pycosim -m pytest hb-pytorch/hammerblade/torch/tests/test_cat.py -vs` and tests passed

perfect!

yodada:
This PR partially implements `torch._cat` -- currently we only support `_cat` with `dim=0`

PR #153, opened 2020-08-31
Line trace update

kr397:
Added `pc` and `full` modes. Changed single character stall codes to make them unique.


yodada:
Nice work!

I was wondering if you could include a simple change to `print_trace.py` here

https://github.com/cornell-brg/hb-pytorch/blob/b4bfcade801c7da88dfffe2f91d5e8f5fd68eb4b/hammerblade/scripts/line_trace/print_trace.py#L49-L54
https://github.com/cornell-brg/hb-pytorch/blob/b4bfcade801c7da88dfffe2f91d5e8f5fd68eb4b/hammerblade/scripts/line_trace/print_trace.py#L65-L73

Basically this allows you to specify multiple intervals. For instance, you can do `--lo 1-7 --lo 9-15` to show the first 2 rows except the 2 tiles on column 0

kr397:
Thanks! I've also added the multiple interval support as you described, let me know if there is anything else

yodada (approved):
LGTM!

PR #154, opened 2020-09-04
Nbload attribute update; Use link script generator

vb000:
- Added prefixes to nbload related keywords.
- Updated build to use bsg_manycore's link script generator instead of a custom link script.


yodada (approved):
LGTM! Thanks for the patch!

vb000:
A little bit on the linker script... we switched to python based link script generator in manycore a while ago. But, I had added a custom script for hb-pytorch until the linker script generator is stable, and also to be able to make quick fixes for pytorch. Then generator is stable is now and includes updates to exclude unnecessary static data in dmem, so we use that now.

PR #156, opened 2020-10-08
bitwise and kernel

Janicewei:
bitwise and kernel with two tensor inputs, only accept integer and boolean

type checks needed

### DIFF
--- a/aten/src/ATen/native/hammerblade/And.cpp
+++ b/aten/src/ATen/native/hammerblade/And.cpp
@@ -6,17 +6,14 @@
 #include <ATen/native/hammerblade/Offload.h>
 
 namespace at { namespace native {
-namespace {
 
-static void and_kernel_hb(TensorIterator& iter) {
-  AT_DISPATCH_FLOAT_AND_INTS(iter.dtype(), "and_hb", [&]() {
-      offload_iterator_reduce_op_impl<scalar_t>(iter, "tensorlib_and");
-      });
+Tensor and_kernel_hb(const Tensor& self, const Tensor& other) {
+  TORCH_CHECK(self.numel() == other.numel(), "The size of two tensors should match.");
+  // TORCH_CHECK(self.scalar_type() == ScalarType::Int || self.scalar_type() == ScalarType::Bool, "HammerBlade and is implemented for Int and Bool only");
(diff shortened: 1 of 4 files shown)


yodada:
I think you mentioned that bitwise and works for booleans as well, right? We probably need a test for booleans as well.

aten/src/ATen/native/hammerblade/And.cpp
```diff
@@ -0,0 +1,19 @@
...
+#include <ATen/native/TensorIterator.h>
+#include <ATen/native/ReduceOps.h>
+#include <ATen/native/hammerblade/Offload.h>
+
+namespace at { namespace native {
+
+Tensor and_kernel_hb(const Tensor& self, const Tensor& other) {
+  TORCH_CHECK(self.numel() == other.numel(), "The size of two tensors should match.");
+  // TORCH_CHECK(self.scalar_type() == ScalarType::Int || self.scalar_type() == ScalarType::Bool, "HammerBlade and is implemented for Int and Bool only");
+  // TORCH_CHECK(other.scalar_type() == ScalarType::Int || other.scalar_type() == ScalarType::Bool, "HammerBlade and is implemented for Int and Bool only");
```
yodada: If you want to check the data type of tensors, the canonical way is like this
```
./aten/src/ATen/native/BinaryOps.cpp:94:  TORCH_CHECK(self.scalar_type() != kBool || other.scalar_type() != kBool,
```

yodada: Actually, with boolean data the emulator segfaults when you want to quit ... I kind of think the reason is your kernel uses 32bit int internally, and I'm not too sure about the size of booleans

yodada:
```
>>> x = torch.ones(10, dtype=torch.bool)
>>> x.element_size()
1
```
yeah so a booleans is a single byte in pytorch

Janicewei: > Actually, with boolean data the emulator segfaults when you want to quit ... I kind of think the reason is your kernel uses 32bit int internally, and I'm not too sure about the size of booleans

how can I get around with the segfault?

Janicewei: I made two functions in kernel_and (one for int and one for bool). If type is bool, do bool fun; if type is int do int fun. But I still have segfault for bool type.

PR #157, opened 2020-10-08
Profiler bug fix

yodada:
This PR fixes a bug which may lead to crash in PyTorch
  - What happened -- hb_mc trimming calls are logged even outside ROI, with an empty stack record. So when you do profiling, and then run any hammerblade operators, dumping profiling data leads to an invalid stack pop.
 - The bug is fixed, and a direct test is added.

Now per operation profiling runs only if `HB_REDISPATCH` is set -- to reduce native profiling overhead


vb000 (approved):
Cool! I remember facing with this issue when native execution is multi-threaded.

PR #158, opened 2020-10-08
kernel bitwise_or

Janicewei:
bitwise_or with two tensor inputs, supports int and bool types

### DIFF
--- a/aten/src/ATen/native/hammerblade/Or.cpp
+++ b/aten/src/ATen/native/hammerblade/Or.cpp
@@ -9,10 +9,15 @@ namespace at { namespace native {
 
 Tensor or_kernel_hb(const Tensor& self, const Tensor& other) {
   TORCH_CHECK(self.numel() == other.numel(), "The size of two tensors should match.");
-  // TORCH_CHECK(self.scalar_type() == ScalarType::Int || self.scalar_type() == ScalarType::Bool, "HammerBlade or is implemented for Int and Bool only");
-  // TORCH_CHECK(other.scalar_type() == ScalarType::Int || other.scalar_type() == ScalarType::Bool, "HammerBlade or is implemented for Int and Bool only");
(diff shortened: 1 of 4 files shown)


aten/src/ATen/native/hammerblade/Or.cpp
```diff
@@ -0,0 +1,19 @@
...
+#include <ATen/native/TensorIterator.h>
+#include <ATen/native/ReduceOps.h>
+#include <ATen/native/hammerblade/Offload.h>
+
+namespace at { namespace native {
+
+Tensor or_kernel_hb(const Tensor& self, const Tensor& other) {
+  TORCH_CHECK(self.numel() == other.numel(), "The size of two tensors should match.");
+  // TORCH_CHECK(self.scalar_type() == ScalarType::Int || self.scalar_type() == ScalarType::Bool, "HammerBlade or is implemented for Int and Bool only");
+  // TORCH_CHECK(other.scalar_type() == ScalarType::Int || other.scalar_type() == ScalarType::Bool, "HammerBlade or is implemented for Int and Bool only");
```
yodada: Same as `and`, I think if you do `self.scalar_type() == kInt` it will work
Maybe you need `at::kInt` ...

hammerblade/torch/kernel/kernel_or.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include <kernel_common.hpp>
+#include <cstdint>
+
+extern "C" {
+
+  __attribute__ ((noinline))  int tensorlib_or(
+          hb_tensor_t* t0_p,
+          hb_tensor_t* t1_p,
+          hb_tensor_t* t2_p) {
+    auto res = HBTensor<int>(t0_p);
```
yodada, line 17: Similar to `and`, we need to handle booleans and test for it

PR #166, opened 2021-03-15
Sinkhorn updates

neiladit:
These are the changes:

- Cdist kernel is offloaded to Hammerblade using `cdist_stub` in `Distance.cpp`
- Sampled dense-dense multiplication kernel (`sddtmm`) on CPU has been updated to operate on CSR format matrices which improves the latency. It is now equivalent to the `dstmm` kernel which has the same FLOPs.
- `sddtmm` kernel on Hammerblade is using `empty` instead of `zeros` to initialize arrays which improves timings since `zeros` calls `fill` kernel on Hammerblade.
- Some test files for sinkhorn borrowed from Andrew and Adrian's branch and altered to fit in cdist kernel as well.


zzy82518996:
I think we need to clean up the sinkhorn_tests folder, and only move test_sinkhorn.py to hammerblade/torch/test/ path. The other part looks good to me !

PR #175, opened 2021-08-18
sigmoid_backward, mse, mse_backward, power (tensor-scalar) operations

nsmlzl:
HammerBlade kernels for following operations:
* sigmoid_backward
* mse_loss
* mse_backward
* power (Tensor base, scalar exp)

Power operation was added, since the Python implementation of [mse_loss for targets requiring gradients](https://github.com/cornell-brg/hb-pytorch/blob/084cf4d483e4adc2451d6875eb7cafd91d7ea47b/torch/nn/functional.py#L2211) requires a square operation.

Kernels are implemented similar to CPU implementation and pass tests in emulation and cosimulation.


zzy82518996 (commented):
Looks good to me. Just a minor thing, could you remove the prints inside the test_mse_backward.py, test_pow_tensor_scalar.py and test_mse_loss.py ? Thanks.

nsmlzl:
I removed the print statements with the latest commit.

PR #176, opened 2021-09-27
Where operator

nsmlzl:
Added where operator implementation for hammerblade;

Since the input tensors of the where operator are not all floats (conditional tensor is of type bool or uint8_t), some changes in the hb_tiled_for.hpp file were necessary. Previously, the hb_tiled_foreach functions expected all tensors to be of same type. Now, the type of each tensor is saved in the *data_type* definition of the hb_tensor class. the hb_tiled_foreach functions utilizes this definition.

Since many other operators utilize the hb_tiled_foreach procedure, the changes in the hb_tiled_for.hpp file should be checked carefully. For me it compiles fine and most pytests run successfully.

### DIFF
--- a/aten/src/ATen/native/hammerblade/Where.cpp
+++ b/aten/src/ATen/native/hammerblade/Where.cpp
@@ -1,0 +1,28 @@
+#include <ATen/Dispatch.h>
+#include <ATen/hammerblade/HammerBladeContext.h>
+#include <ATen/native/TensorIterator.h>
+#include <ATen/native/hammerblade/Offload.h>
+
+namespace at { namespace native {
+
+Tensor _s_where_hb(const Tensor& condition, const Tensor& self, const Tensor& other) {
+    TORCH_CHECK(self.dtype() == other.dtype(), "expected scalar type ", self.dtype(), " but found ", other.dtype());
+    Tensor result = at::empty(self.sizes(), self.options());
+
+    auto iter = at::TensorIterator();
+    iter.set_check_mem_overlap(true);
+    iter.add_output(result);
+    iter.add_input(condition);
+    iter.add_input(self);
+    iter.add_input(other);
+    iter.dont_compute_common_dtype();
+    iter.build();
+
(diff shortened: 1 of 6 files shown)


aten/src/ATen/native/hammerblade/Where.cpp
```diff
@@ -0,0 +1,34 @@
...
+    iter.set_check_mem_overlap(true);
+    iter.add_output(result);
+    iter.add_input(condition);
+    iter.add_input(self);
+    iter.add_input(other);
+    iter.dont_compute_common_dtype();
+    iter.build();
+
+    if (condition.scalar_type() == at::ScalarType::Byte) {
+        AT_DISPATCH_FLOAT_TYPE_ONLY(iter.dtype(), "where_byte_hb", [&](){
```
nsmlzl, line 22: @zzy82518996 Regarding our discussion of 'AT_DISPATCH_FLOAT_TYPE_ONLY':
I think this works, because it only checks the type of the first tensor (here of output, which is of type float): 
[TensorIterator::dtype()](https://github.com/cornell-brg/hb-pytorch/blob/084cf4d483e4adc2451d6875eb7cafd91d7ea47b/aten/src/ATen/native/TensorIterator.h#L200)

Is this behaviour in general acceptable for you? 

Surely, I can quickly change it to 'AT_DISPATCH_ALL_TYPES' if you would prefer.

hammerblade/torch/kernel/hb_tiled_for.hpp
```diff
@@ -128,13 +128,13 @@ __attribute__((noinline)) void hb_tiled_foreach_impl(
 }
 
 // Unary
-template<typename scalar_t, typename F, typename... P>
+template<typename scalar1_t, typename scalar2_t, typename F, typename... P>
```
zzy82518996, line 131: I am a little bit worried about this part, since this function is called by every other element-wise kernels. Can you also check whether it still works for the other element-wise kernels ?

PR #181, opened 2022-01-28
Fix emulation mode for index-add operation

nsmlzl:
Use cpp std::mutex in emulation mode, since mcs-mutex from bsg_manycore_lib does not work in emulation.
No change in cosimulation mode (continue to use mcs-mutex).


zzy82518996:
Cool, looks good to me. Last thing, could you make the code style consistent so that it can pass the verification when merging into master. All the python file should have indent 4, and all the C++ file should have indent 2. Thanks.

PR #182, opened 2022-02-01
Fix formatting

nsmlzl:
Fix formatting for:
* sigmoid_backward
* mse
* mse_backward
* power
* where
* index_add


zzy82518996:
Looks good to me !

